Issue #9, opened 2022-11-07
Reduce default content in GUI


coleramos425:
In the standalone GUI, when no filters are applied
```
omniperf analyze -p workloads/sample/mi200/ --gui
````
the HTML page will load data for every single metric and chart. To reduce loading time and compute, only high-level sections should be displayed:

- Top Kernels
- Speed-of-Light
- Memory Chart

The rest of the information can be displayed when kernel or dispatch filters are applied. Which will significantly decrease the compute required to generate results.

PR #15, opened 2022-11-09
Improve GUI Useage

coleramos425:
Few changes to enhance usage in the standalone GUI:

1. **Only show basic metrics when no filters have been applied**
(This will reduce processing time and as soon as the user applies a filter, either via the GUI or CLI, we show the full detailed output)

2. **Implement a basic spinner to indicate loading**
(This will eventually be replaced with a progress bar, but we're adding this in the meantime)


coleramos425:
Closes #9

PR #35, opened 2022-11-17
Ubuntu roofline binary isn't installed

coleramos425:
Add code to include Ubuntu 20.04 roofline binary in packaging. 

- Modified CMakeLists


koomie (approved):
Looks good to me.

PR #57, opened 2022-12-16
Add better error detection when local ROCM installation is incomplete.

### LINKED ISSUE #56: Add better error detection when ROCm install is incomplete
Omniperf presently relies on the .info directory included with normal ROCm install to determine versioning information. If this directory is missing (say, due to incomplete ROCm install), the user will encounter runtime errors.

Improve the error message in this case to indicate the ROCm installation is incomplete.

koomie:
Closes #56.

Signed-off-by: Karl W. Schulz <karl.schulz@amd.com>

### DIFF
--- a/src/utils/specs.py
+++ b/src/utils/specs.py
@@ -24,6 +24,7 @@
 
 import os
 import re
+import sys
 import socket
 import subprocess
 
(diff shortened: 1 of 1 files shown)


jrmadsen:
FYI, the .info files don’t necessarily exist if ROCm is installed via Spack. In omnitrace, when I can’t find them, I figure them out by finding the HIP version reported by cmake but in your case, it might be worth considering just letting them specify the ROCm version in the env as a last resort.

koomie:
Thanks for the suggestion. Seems like we should try to get the spack recipe updated if that's the case. In the meantime, I have done as you suggested and allowed an env variable to override if we can't detect from .info/version.

PR #88, opened 2023-02-13
remove unused file

koomie:
@coleramos425 , I don't believe we are using `cmake/miperf.sh.in` anymore. Can you confirm and if so, can we consider removal?


coleramos425:
Yes, let's get that removed to avoid confusion. Thanks for catching that.

Issue #75, opened 2023-01-25
PMC metric not found


aosewski:
Hi,

I'm trying to profile a program on MI100 GPU and got following error:
```
ERROR: rocprofiler_open(), Initialize(), input metric 'TCP_GATE_EN1_sum' is not found
```
Similarly the metric `SQ_IFETCH` is also not found. 
I'm using ROCm 5.3.0 and I can't find those metrics for MI100 in rocprofile https://github.com/ROCm-Developer-Tools/rocprofiler/blob/rocm-5.3.0/test/tool/gfx_metrics.xml#L80

I'm using version 1.0.6.

[EDIT]
I see that such _expressions_ are defined here: https://github.com/ROCm-Developer-Tools/rocprofiler/blob/amd-master/test/tool/metrics.xml#L127 However this one and a lot of others are defined only for MI200. And you are using them for MI100: https://github.com/AMDResearch/omniperf/blob/main/src/perfmon_pub/mi100/pmc_tcp_perf.txt

coleramos425:
Thank you for reporting this discrepancy @aosewski. It's possible there are some counters that may have slipped into MI100 configs we overlooked.

Let me cross-check our listings and I'll report my findings in this thread.

coleramos425:
Update: I've confirmed this issue on ROCm 5.3.x and 5.4.x releases. Looking into possible solutions to include in our v1.0.7 release

coleramos425:
Fix has been merged into `dev` branch. Closing issue.

PR #91, opened 2023-02-15
Custom metrics for mi100

coleramos425:
Regarding the report of missing metrics in mi100 (#75) default rocprof configuration files. This PR provides a metric file for gfx908 and overrides the incomplete default metrics.

Logic stays the same unless gfx908 is detected.


coleramos425:
Approving. Verified on MI100 hardware.

PR #116, opened 2023-04-19
Update requirements.txt to include setuptools

dipietrantonio:
On some systems setuptools (needed by dash-bootstrap-components) is not installed by default. Weirdly enough, it is not installed as dependency of Dash.


coleramos425:
Thank you for the submission @dipietrantonio. We'll get this merged into our dev branch so it gets included in our next release (v1.0.8)

P.S. Ignoring the two failing tests. Fixes to these have been applied recently.

PR #122, opened 2023-05-03
CLI: fix decimal doesn't work for single run

feizheng10:


coleramos425:
Thanks Fei! We'll ignore the 2 failing CI tests which are just checkout-related errors on our end.

PR #124, opened 2023-05-03
refactoring saved df files

feizheng10:
Save analysis dataframes in that loading function with verbose actually is not a bad idea! 
However, there are some reasons to refactoring the code a bit:
* Writing to any dir without asking user is not a good practice(My current implementation to write TopStats there is a tmp solution, it is not good and need to be fixed in proper way ). 
* Implement it in tty will guarantee to save whatever user saw in cli. It is hard to imagine users want to save dfs in gui when loading(may be a bit confused when changing filter).
* It is easy to retrieve table title info in tty and put it into file name.

### DIFF
--- a/src/parser.py
+++ b/src/parser.py
@@ -469,6 +469,12 @@ def parse(my_parser):
         help="\t\tSpecify the decimal to display. (DEFAULT: 2)",
     )
     analyze_group.add_argument(
+        "--save-dfs",
+        dest="df_file_dir",
+        metavar="",
+        help="\t\tSpecify the dirctory to save analysis dataframe csv files.",
+    )
+    analyze_group.add_argument(
         "--cols",
         type=int,
         dest="cols",
--- a/src/omniperf_analyze/omniperf_analyze.py
+++ b/src/omniperf_analyze/omniperf_analyze.py
@@ -209,6 +209,7 @@ def run_cli(args, runs):
             runs,
             archConfigs["gfx90a"],
             output,
+            args.df_file_dir,
             args.decimal,
             args.time_unit,
             args.cols,
--- a/src/omniperf_analyze/utils/parser.py
+++ b/src/omniperf_analyze/utils/parser.py
@@ -718,22 +718,6 @@ def load_table_data(workload, dir, is_gui, debug, verbose):
         debug,
     )
 
-    # Save workload
-    name = "saved_analysis"
-    out_path = os.path.join(dir, name)
-    try:
-        os.mkdir(out_path)
-        if verbose >= 1:
-            print("Created a Saved Analysis folder")
-    except OSError as error:
-        if verbose >= 1:
-            print("Saved Analysis folder exists")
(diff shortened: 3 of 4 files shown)


coleramos425:
Tip for formatting test:
```shell
python3 -m pip install black
cd omniperf/
black --diff --check . # check all files for any formatting discrepancies
black . # re-format files which don't comply
```

feizheng10:
> ```shell
> black --diff --check .
> ```

Thanks! As long as we black with the same format style:)

feizheng10:
> > ```shell
> > black --diff --check .
> > ```
> 
> Thanks! As long as we black with the same format style:)

Just check some discussion back a while ago...
I would suggest to use yapf instead of black. The reason is: a few ROCm components have being evaluated python formatting solutions and we are all on the same page that yapf default formatting is much better than black, even though it is not perfect. 


More details you might follow:

yapf installation: sudo apt install yapf
	
yapf version >= 0.32.0
	
For your local file formatting: yapf --style pep8 -ip your_file
	
CI verification: find . -iname '*.py' | grep -v 'build/' | xargs -n 1 -P 1 -I{} -t sh -c 'yapf -ip --style pep8 {}
	
Disable/enable syntax:
	
	# yapf: disable
	...
	
	# yapf: enable

coleramos425:
> I would suggest to use yapf instead of black.

Thanks for the suggestion. Omniperf uses _black_ format to remain consistent with the rest of the Omni* toolset (i.e. Omnitrace). I'll need to evaluate performance differences

src/parser.py
```diff
@@ -468,6 +468,12 @@ def parse(my_parser):
         default=2,
         help="\t\tSpecify the decimal to display. (DEFAULT: 2)",
     )
+    analyze_group.add_argument(
+        "--save-dfs",
+        dest="df_file_dir",
+        metavar="",
+        help="\t\tSpecify the dirctory to save analysis dataframe csv files.",
+    )
```
coleramos425, line 476: I agree with your logic. In the future saving analysis will be helpful, but only when logic is available to to use this saved analysis as cached data to reduce processing time.

Until then, I find the `--save-dfs` is not needed since we already have the `--output` flag to save output data. In the interest of keeping CLI options simple, I'm commenting out.

```suggestion
    # analyze_group.add_argument(
    #    "--save-dfs",
    #    dest="df_file_dir",
    #    metavar="",
    #    help="\t\tSpecify the dirctory to save analysis dataframe csv files.",
    # )
```

feizheng10: The scenario of --output is different.  --output saves only the text report, which can be shared to others as overview.
--save-dfs save all the dfs in csv, which can be used for further analysis or other visualization. @arghdos, does this match your expectation?

arghdos: Yes -- I need the raw dataframes quite often, and dumping the pared output (with all of the crazy unicode chars) is not really a substitute.  IMO, these can co-exist.

PR #125, opened 2023-05-08
Manually join rocprof runs

coleramos425:
In an attempt to address kernel merging issues with default rocprofiler implementation (i.e. **$ROCPROF_DIR/libexec/rocprofiler/tblextr.py**), we've created [join_prof()](https://github.com/AMDResearch/omniperf/blob/298271e1d083c75c0da11dea120a0f31fac553eb/src/utils/perfagg.py#L90).

-  [pmc_perf_split()](https://github.com/AMDResearch/omniperf/blob/298271e1d083c75c0da11dea120a0f31fac553eb/src/utils/perfagg.py#L146) separates `pmc_perf.txt` into 17 input files (one for each line)

- 17 separate output files (pmc_perf_*.csv) are generated by rocprof and fed into [join_prof()](https://github.com/AMDResearch/omniperf/blob/298271e1d083c75c0da11dea120a0f31fac553eb/src/utils/perfagg.py#L90)

- [join_prof()](https://github.com/AMDResearch/omniperf/blob/298271e1d083c75c0da11dea120a0f31fac553eb/src/utils/perfagg.py#L90) merges all 17 files into `pmc_perf.csv` by combining counters across kernels and averaging timestamps


coleramos425:
Everything is looking good in my tests. Only issue I've noticed is that `gpu-id`, which is used in analyze mode (i.e. for gpu-id based filtering), is not accounted for at the moment
https://github.com/AMDResearch/omniperf/blob/8c173446d2909db48c6cfe024ab54205fbfef1e8/src/utils/perfagg.py#L107-L132

If `gpu-id` will always be consistent across all runs then it's simple to merge. However, I'm thinking there might be some edge cases where this isn't true... (i.e. MPI runs?) which may introduce some complications. Any thoughts @arghdos ?

arghdos:
>However, I'm thinking there might be some edge cases where this isn't true... (i.e. MPI runs?) which may introduce some complications. Any thoughts @arghdos ?

You're probably right that there could be some cases where different MPI runs target different GPUs per process.  We could probably code up some strategies the user can select from for joing such an app, but in the short term I'm less concerned about that.

src/utils/perfagg.py
```diff
@@ -86,6 +87,121 @@
...
+# joins disparate runs less dumbly than rocprof
+def join_prof(workload_dir, out):
+    files = glob.glob(workload_dir + "/" + "pmc_perf_*.csv")
+    df = None
+
+    for i, file in enumerate(files):
+        # _df = parse_rocprof_kernels(file)
+        _df = pd.read_csv(file)
+        key = _df.groupby("KernelName").cumcount()
+        _df["key"] = _df.KernelName + " - " + key.astype(str)
```
arghdos, line 99: We probably want two versions of this key calculation here:

1. What we have here, essentially "unique kernel name dispatches"
2. Adding another " - " + 'grd' to get the total grid size, that gives us ~ what we discussed last time (join by kernel grid size + name + dispatch id)

Ideally, we can switch between the two (not sure what the default should be, perhaps the 'grid' strategy?) via a command line arg

arghdos (changes requested):

coleramos425:
I've redefined "boring duplicates" such that `gpu-id` required in filtering, along with `wgr`, `vgpr`, `sgpr`, `lds`, `scr`, `grd` required in wavefront launch are no longer ignored.
https://github.com/AMDResearch/omniperf/blob/444102d92726baf286d1cb321fc2eb3f1bc30ee3/src/utils/perfagg.py#L119-L144

coleramos425:
Note we also have a `--join-type` option
https://github.com/AMDResearch/omniperf/blob/444102d92726baf286d1cb321fc2eb3f1bc30ee3/src/parser.py#L185-L192

https://github.com/AMDResearch/omniperf/blob/444102d92726baf286d1cb321fc2eb3f1bc30ee3/src/utils/perfagg.py#L101-L108

src/utils/perfagg.py
```diff
@@ -86,6 +87,133 @@
...
+    df = df.drop(columns=ekeys)
+    df["BeginNs"] = beginNs
+    df["EndNs"] = endNs
+    # finally, join the drop key
+    df = df.drop(columns=["key"])
+    # and save to file
+    df.to_csv(out, index=False)
+    # and delete old file(s)
+    for file in files:
+        os.remove(file)
```
arghdos, line 182: It occurs to me, we may want to leave the old files in place in VERBOSE mode (I think is the name? can't remember -- the one you added the other week that keeps more logging information around)

arghdos (approved):
This looks good to me, minor comment about "perhaps we wanna save the individual files" for logging purposes.

Thanks @coleramos425 !

src/utils/perfagg.py
```diff
@@ -86,6 +87,133 @@
...
+    if out == None:
+        out = workload_dir + "/pmc_perf.csv"
+    files = glob.glob(workload_dir + "/" + "pmc_perf_*.csv")
+    df = None
+
+    for i, file in enumerate(files):
+        # _df = parse_rocprof_kernels(file)
+
+        _df = pd.read_csv(file)
+        key = _df.groupby("KernelName").cumcount()
```
arghdos, line 102: I actually thought of something overnight -- when we are trying to join via 'grid', we need to actually generate the key based on the groupby(("KernelName", "grd")), such that we count unique dispatches of each kernel + grid combination.

I think that actually just means we need to move the `join_type` up to this line, to do like:

```
if join_type == "kernel":
  key = _df.groupby("KernelName").cumcount()
elif join_type == "grid":
  _df.groupby(("KernelName", "grd")).cumcount()
else:
  <error>
```

Then I think we just continue as below

src/utils/perfagg.py
```diff
@@ -86,6 +87,133 @@
...
+            for k in df.keys()
+            if not any(
+                check in k
+                for check in [
+                    # "gpu",
+                    "queue-id",
+                    "queue-index",
+                    "pid",
+                    "tid",
+                    # "grd",
```
arghdos, line 132: Do we do any checking that e.g., the vgpr counts match between joins?
That might be a good warning to throw (e.g., check if we get any of the suffixes above, and if so warn the user they may want to use a different strategy)

src/utils/perfagg.py
coleramos425: https://github.com/AMDResearch/omniperf/blob/44edef4635c18ee3f2bb764f7078db05dfbbecd6/src/utils/perfagg.py#L211-L214

src/utils/perfagg.py
coleramos425: https://github.com/AMDResearch/omniperf/blob/44edef4635c18ee3f2bb764f7078db05dfbbecd6/src/utils/perfagg.py#L104-L111

src/utils/perfagg.py
coleramos425: https://github.com/AMDResearch/omniperf/blob/44edef4635c18ee3f2bb764f7078db05dfbbecd6/src/utils/perfagg.py#L121-L140

coleramos425:
Thanks for the review. Looks good to me. Merging this into dev for others to try out.

PR #132, opened 2023-05-24
Fix merge on grid

arghdos:
This fixes an issue reported by a user trying the new merge functionality on Pele.

The issue (essentially) was that when using the 'grid' join type, we were:

- creating the cumulative count (the 'key' series) by coind a cumcount() on `_df.groupby(["KernelName", "grd"])`: https://github.com/AMDResearch/omniperf/blob/c5a19ca06a879302ba183aa08033849ba448a407/src/utils/perfagg.py#L108
- However, when actually creating the key to join on (https://github.com/AMDResearch/omniperf/blob/c5a19ca06a879302ba183aa08033849ba448a407/src/utils/perfagg.py#L113) we were not including the "grd" as part of the key. 

The result was that we would have many duplicate keys, and on the merge the data-frame would ~ double in size.  Rinse and repeat a dozen times or so, and we would observe a OOM crash.

This fixes this issue by making the "key" entry we join on match the cumulative count we create.  We may want to think of more robust error detection here though.  Perhaps we could experiment with the "one_to_one" validation scheme in `pandas`?

### DIFF
--- a/src/utils/perfagg.py
+++ b/src/utils/perfagg.py
@@ -122,7 +122,7 @@ def join_prof(workload_dir, join_type, log_file, verbose, out=None):
     duplicate_cols = {
         "gpu": [col for col in df.columns if "gpu" in col],
         "grd": [col for col in df.columns if "grd" in col],
-        "wpr": [col for col in df.columns if "wgr" in col],
+        "wgr": [col for col in df.columns if "wgr" in col],
         "lds": [col for col in df.columns if "lds" in col],
         "scr": [col for col in df.columns if "scr" in col],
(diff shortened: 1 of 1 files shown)


arghdos:
@PaulMullowney -- can you try this?

PaulMullowney:
> @PaulMullowney -- can you try this?

yup.

coleramos425:
@arghdos to comply with Python formatting could you:
```
python3 -m pip install black
cd home/arghdos/omniperf-dir/
black .
git commit -s -m "Begrudgingly comply to formatting"
```
Thanks :)

Issue #144, opened 2023-06-30
Only evaluate metrics in analyze for architectures we're actually going to use!


arghdos:
We know that the python parsing of metrics has high overhead on each analysis / GUI run.
However, I did _not_ know that we are parsing&compiling metrics for all available `gfx90X` architectures each analyze run.
I discovered this while modifying some of the `yaml` files, wherein I discovered that I had a syntax error in one of my modified config files.
I fixed it in `gfx90a` only, to test the change, but was quite surprised when I kept hitting the same syntax error!

Backtracing in `pdb` lead me to:

https://github.com/AMDResearch/omniperf/blob/a346db7646b0a935f4cac51d131b4a585f065c05/src/omniperf_analyze/omniperf_analyze.py#L97

where-in I realized, we're compiling for all architectures:

```
p archConfigs.keys()
dict_keys(['gfx906', 'gfx908', 'gfx90a'])
```

A quick optimization (that will lower the overhead of analyze by ~3x) is to detect which architecture we're going to analyze (at the very least, we have access to the `miXXX` part of the `--path` the user passed in, but it's also in the `sysinfo.csv` file in said path), and only parse&compile metrics for that arch.

coleramos425:
Things are a bit more complicated since we support baseline comparison functionality in CLI. If I'm conducting an intra-soc comparison we'd need ArchConfigs to be set for each architecture, i.e.
```console
$ omniperf analyze -p workloads/test_1/mi200/ -p workloads/test_2/mi100/
```
For now, I'm just detecting the "newest" architecture loaded in ArchConfigs and using this to form the baseline comparison data tables. This should work in the short term considering newer archs should contain a superset of all tables avail in older archs.

May want to revisit this in the ["Omniperf Revamp"](https://github.com/AMDResearch/omniperf/discussions/153)...

arghdos:
[AMD Official Use Only - General]

Really? I had no idea we supported that!
Neat!

I think we could do the same sort of conditional "we see mi100 and mi200 in the workloads thing, so only eval those two"?
But I am fuzzy on how all this mapping works

________________________________

coleramos425:
Merged into `dev`. Closing issue.

PR #152, opened 2023-07-19
Only load ArchConfigs for detected SoC

JoseSantosAMD:
Addresses issue #144


feizheng10:
Sort of, related to this, it is better to fix hack code like https://github.com/AMDResearch/omniperf/blob/ed31b8a988b0fde6a5e11bc949417c82c6db1abc/src/omniperf_analyze/omniperf_analyze.py#L205 as well

coleramos425:
Rejecting PR. Doesn't take into account multiple paths which could have different SoCs (i.e. intra SoC comparison) -- this case breaks your logic. For example,

```console
omniperf analyze -p workloads/test_1/mi200 -p workloads/test_2/mi100
```

I'll take this PR and submit a re-worked version

PR #156, opened 2023-08-02
Bw smart units

JoseSantosAMD:


coleramos425:
Let's also make sure we're running a test scenario with `--gui` for this PR.

coleramos425:
Looks good. I confirmed this worked using nbody workload to trigger low SOL results

PR #161, opened 2023-08-15
Bw smart units new

### LINKED ISSUE #145: Smarter units for bandwidths
It would be nice if we could get units for bandwidth values that scale with the size.  E.g., if we have > a GB/s, label it GB/s, or MB/s or KB/s, etc.

Not sure how possible this actually is in the current design, but would improve readability significantly.

cc: @feizheng10

JoseSantosAMD:
Closes #145

### DIFF
--- a/src/omniperf_analyze/configs/gfx906/0200_system-speed-of-light.yaml
+++ b/src/omniperf_analyze/configs/gfx906/0200_system-speed-of-light.yaml
@@ -107,7 +107,7 @@ Panel Config:
           LDS BW:
             value: AVG(((((SQ_LDS_IDX_ACTIVE - SQ_LDS_BANK_CONFLICT) * 4) * TO_INT($LDSBanks))
               / (EndNs - BeginNs)))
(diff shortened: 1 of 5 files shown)


coleramos425:
The changes made don't account for the baseline comparison edge case, i.e.
```console
$ omniperf analyze -p workloads/mix/mi200/ -p workloads/vcopy/mi200/ -b 2.1.22
 Index   │ Metric                    │ Value   │ Value           │ Unit             │ Peak     │ Peak            │ PoP   │ PoP             │ 
2.1.22  │ L2-Fabric Read BW         │ 50.0    │ 0.14 (-99.72%)  │ Mb/s             │ 1638.4   │ 1638.4 (0.0%)   │ 0.0   │ 0.01 (inf%)  
```
This example illustrates the fact that the "smart unit" will be determined by the first workload, but the second workload is still in unchanged Gb/s form. Please revisit this.

coleramos425:
Stale branch

PR #166, opened 2023-08-22
Added additional arguments for application prof

WillTrojak:
Added the feature to profile, that additional arguments can be given that are passed to rocprof. This is mainly in response to https://github.com/AMDResearch/omniperf/issues/164 and to allow trace start to be set, but it seemed that other options may be useful and this was the best solution.

### DIFF
--- a/src/omniperf
+++ b/src/omniperf
@@ -270,7 +270,7 @@ def roof_setup(args, my_parser, VER):
     # Do we need a new directory for roofline?
     if not os.path.isdir(args.path):
         os.makedirs(args.path)
-    
+
     # Does roof data exist?
     print("Checking for roofline.csv in ", args.path)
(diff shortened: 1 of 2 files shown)


coleramos425:
Thanks for your contribution @WillTrojak. As stated in our [CONTRRIBUTING.md](https://github.com/AMDResearch/omniperf/blob/main/CONTRIBUTING.md) guide, we require users sign-off on commits

> Omniperf requires new commits to include a "Signed-off-by" token in the commit message (typically enabled via the git commit -s option), indicating your agreement to the projects's [Developer's Certificate of Origin](https://developercertificate.org/) and compatability with the project [LICENSE](https://github.com/AMDResearch/omniperf/blob/main/LICENSE):

Before we can accept, please follow instructions here:
image

PR #167, opened 2023-08-23
U22 dockerfile

JoseSantosAMD:


.github/workflows/ubuntu-focal.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     # The type of runner that the job will run on
     runs-on: ubuntu-latest
     container:
-       image: rocm/dev-ubuntu-22.04:5.3
+       image: rocm/dev-ubuntu-22.04:5.3.3-complete
```
coleramos425, line 21: This is the ubuntu-focal workflow, it should be version 20.04, not 22.04

coleramos425 (changes requested):

PR #197, opened 2023-10-31
Introduce a cmake option to enable python code coverage.

koomie:


coleramos425:
This comment is coming a bit late but per my observation in the latest PR (below)

https://github.com/AMDResearch/omniperf/pull/204#issuecomment-1791575753

It occurs to me that instead of keeping these directories static, we may create a separate build step so that if Python dependencies change we don't need to update those libs manually

https://github.com/AMDResearch/omniperf/blob/df4d508a03e114d89c47b4e67888dc4273bcff3c/.github/workflows/mi100.yml#L19-L21

PR #200, opened 2023-10-31
Initial overhaul of Profile mode

coleramos425:
My initial PR includes a redesign of our "Profile mode" code. From a UX perspective, nothing should be different. All your usual profile commands will run just as before.

Under the hood, I've rearranged the `/src/` directory to reflect some files that have been deleted, renamed, or otherwise integrated into one of our classes as helper methods.

To start, I've just tested these changes on a **Mi200 + rocprofv1 environment**. My plan is to first agree on structuring for this combination, then use that to inform the others.

At a high level, the new design is shown in the "New structure" section below. Any feedback would be highly appreciated - I'm most curious if any improvements can be made to my encapsulation, overall organization and/or logging.

### New structure
image

image

image


feizheng10:
Few quick thoughts:
* I won't call it "metric_configs" because from the design view the configs including both front-end panel/table style and back-end metrics definition/description unless we'd like to separate those in the future.
* Don't see the def of roofline_obj. I don't see any reason it associates to gfx90a_soc(even though we only provide it for gfx90a) .

coleramos425:
> Don't see the def of roofline_obj. I don't see any reason it associates to gfx90a_soc(even though we only provide it for gfx90a)

Thanks for your feedback, @feizheng10. The idea is to, at a high level, put any SoC specific material in these SoC subclasses. Since Roofline (specifically MFMA) isn't available on all hardware types, I decided to make this a property of gfx90a.

Additionally, our roofline calculation may change in future hardware. Instantiating the roofline object at the SoC level gives us the ability to override any parts of the roofline calc from within these SoC subclasses. Let me know if this makes sense

feizheng10:
> > Don't see the def of roofline_obj. I don't see any reason it associates to gfx90a_soc(even though we only provide it for gfx90a)
> 
> Thanks for your feedback, @feizheng10. The idea is to, at a high level, put any SoC specific material in these SoC subclasses. Since Roofline (specifically MFMA) isn't available on all hardware types, I decided to make this a property of gfx90a.
> 
> Additionally, our roofline calculation may change in future hardware. Instantiating the roofline object at the SoC level gives us the ability to override any parts of the roofline calc from within these SoC subclasses. Let me know if this makes sense

I mean, technically, roofline is nothing special other than other ip blocks which should be configed similar way. I'm planing the improve that in my working branch.

feizheng10:
BTW,  prefix "__" for  members in class seems not recommended in PEP8.

koomie:
> BTW, prefix "__" for members in class seems not recommended in PEP8.

I'm a fan of demarcating variables that we intend to treat as private class vars (since Python doesn't have private vars directly).  It doesn't have to be underscores, but I think we should pick a convention for items we are sharing across member functions and try to stick with it.  btw, although it may not be mentioned in PEP8, this common convention is still mentioned in latest python docs, e.g.:

https://docs.python.org/3.11/tutorial/classes.html#private-variables

koomie:
> > BTW, prefix "__" for members in class seems not recommended in PEP8.
> 

Also, in the section on designing for inheritance for PEP8 (https://peps.python.org/pep-0008/#designing-for-inheritance) there is still a reference to the use of leading underscore for non-public attributes (sort of in a backwards way).  It says:

> With this in mind, here are the Pythonic guidelines:
> * Public attributes should have no leading underscores.

So, while there is really no such thing as private attributes, the notion of demarcating those that are intended to be used as private for developers seems to remain.

coleramos425:
Update:
I've just added changes to the vcopy app included with Omniperf. It will now support multi-dispatch and multi-kernel requirements for our CI tests.

Note, there's a hidden option in vcopy to activate multi-kernel (to avoid confusing customer) it's called `--multikernel`

src/omniperf_base.py
```diff
@@ -26,37 +26,47 @@
...
+        # self.__analyze_mode = "cli"
+        # self.__analyze_mode = "webui"
+        
         logging.info("Execution mode = %s" % self.__mode)
    
-
+    def print_graphic(self):
+        """Read program name and log ascii art to terminal.
+        """
+        ascii_art = pyfiglet.figlet_format(config.prog.capitalize())
```
koomie, line 66: Cute but I wonder if we want to introduce another dependency (pyfiglet) just for this? We could embed the results directly perhaps.  No strong opinion, but just thinking about dependency bloat.

src/argparser.py
```diff
@@ -392,31 +381,16 @@ def parse(my_parser):
...
-        help="\t\tList metrics can be customized to analyze on specific arch:\n\t\t   gfx906\n\t\t   gfx908\n\t\t   gfx90a",
-    )
-    analyze_group.add_argument(
-        "-b",
-        "--metric",
-        dest="filter_metrics",
-        metavar="",
-        nargs="+",
-        help="\t\tSpecify IP block/metric id(s) from --list-metrics for filtering.",
+        help="\t\tList all availible metrics for analysis on specified arch:\n\t\t   gfx906\n\t\t   gfx908\n\t\t   gfx90a",
```
koomie, line 393: What do you think of this hard-coded list of arches?

Also, might as well fix the "available" typo.

src/docs/profiling.md
```diff
@@ -42,44 +42,43 @@ The *omniperf* script, availible through the [Omniperf](https://github.com/AMDRe
...
-  -v, --version                   show program's version number and exit
-  -V, --verbose                   Increase output verbosity
+  -v, --version                    show program's version number and exit
+  -V, --verbose                    Increase output verbosity
 
 Profile Options:
   -n , --name                                           Assign a name to workload.
   -p , --path                                           Specify path to save workload.
-                                                        (DEFAULT: /home/colramos/GitHub/omniperf/workloads/<name>)
+                                                        (DEFAULT: /home/colramos/GitHub/omniperf-pub/workloads/<name>)
```
koomie, line 79: Hadn't noticed this before; should probably genericize the DEFAULT path

src/docs/profiling.md
```diff
@@ -113,36 +115,40 @@ The following sample command profiles the *vcopy* workload.
...
-Kernel Selection:  None
-Dispatch Selection:  None
+ROC Profiler: /opt/rocm-5.2.1/bin/rocprof
+Execution mode = profile
+  ___                  _                  __ 
+ / _ \ _ __ ___  _ __ (_)_ __   ___ _ __ / _|
+| | | | '_ ` _ \| '_ \| | '_ \ / _ \ '__| |_ 
+| |_| | | | | | | | | | | |_) |  __/ |  |  _|
+ \___/|_| |_| |_|_| |_|_| .__/ \___|_|  |_|  
+                        |_|
```
koomie, line 125: This is the bit I might suggest to cache directly if want to keep.

src/docs/profiling.md
```diff
@@ -113,36 +115,40 @@ The following sample command profiles the *vcopy* workload.
...
-RPL: output dir '/tmp/rpl_data_230411_165021_26406'
-RPL: result dir '/tmp/rpl_data_230411_165021_26406/input0_results_230411_165021'
-Finished allocating vectors on the CPU
-ROCProfiler: input from "/tmp/rpl_data_230411_165021_26406/input0.xml"
+KernelName verbose: 2
+
+Current input file: /home/colramos/omniperf/workloads/vcopy_test/mi200/perfmon/pmc_perf_2.txt
+RPL: on '231120_142934' from '/opt/rocm-5.2.1' in '/home/colramos/omniperf'
+RPL: profiling '""./vcopy -n 1048576 -b 256""'
+RPL: input file '/home/colramos/omniperf/workloads/vcopy_test/mi200/perfmon/pmc_perf_2.txt'
```
koomie, line 141: more path sanitizing to consider...

src/omniperf
koomie, line 1: That was lovely to get a nice top-level main().

src/omniperf_base.py
coleramos425: Good idea. Done.

koomie (approved):
Thanks for quick updates on pyfiglet and --kernel-verbose.

I'm in favor of going ahead and landing this.

PR #204, opened 2023-11-02
CMake - Load dependency check from requirements.txt

coleramos425:
I noticed this issue when submitting #200 which added a new Python module. I found that in our CMakeLists.txt, we're hardcoding a list of modules for the build script to check for
https://github.com/AMDResearch/omniperf/blob/df4d508a03e114d89c47b4e67888dc4273bcff3c/CMakeLists.txt#L74-L88

This PR changes that logic s.t. we now pull from the requirements.txt file. My CMake experience is  scarce, so I'm open to suggestions if there's a better way to do this

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,20 +72,7 @@ if(DEFINED PYTHON_DEPS)
 endif()
 
 # Check required Python packages
-set(pythonDeps
-    "astunparse"
-    "colorlover"
-    "dash"
-    "matplotlib"
-    "numpy"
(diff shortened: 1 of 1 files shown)


coleramos425:
Looks like pyyaml module is missing, causing mi100 build to fail
image

koomie:
The `pyyaml` error is a but subtle; the issue there is that that package name is `pyyaml`, but the module load on import is `yaml`.  Added some additional logic in cmake config to allow for additional mapping info.

PR #210, opened 2023-12-06
Initial overhaul of Analyze mode

coleramos425:
Please see the changes below for the initial implementation of our revised Analysis mode. Again, this has only been verified on a **Mi200 + rocprofv1 environment**. Once this is merged into the main `2.x` branch we can conduct additional verification against our CI pipeline.

A couple changes to call attention to:

1) Removing hardcoded lists of available archs

https://github.com/AMDResearch/omniperf/blob/4275e15eea222e2bd7f6b18ea2c66b95d4f07904/src/argparser.py#L410-L415

https://github.com/AMDResearch/omniperf/blob/4275e15eea222e2bd7f6b18ea2c66b95d4f07904/src/utils/file_io.py#L237-L245

We've removed instances of hardcoded list, and instead read from archs defined in [src/omniperf_soc/](https://github.com/AMDResearch/omniperf/tree/2.x-dev/src/omniperf_soc)

2) Analyze utilities (i.e. [file_io.py](https://github.com/AMDResearch/omniperf/blob/2.x-dev/src/utils/file_io.py), [parser.py](https://github.com/AMDResearch/omniperf/blob/2.x-dev/src/utils/parser.py), [tty.py](https://github.com/AMDResearch/omniperf/blob/2.x-dev/src/utils/tty.py)) have been moved to the top level utils/ directory


src/argparser.py
```diff
@@ -25,6 +25,27 @@
 import argparse
 import shutil
 import os
+import glob
+
+def load_avail_arch(omniperf_home):
+    '''
+    Load list of avail archs by checking against defined implementations
+    '''
+    matching_files=glob.glob(os.path.join(omniperf_home, 'omniperf_soc', 'soc_*.py'))
```
koomie, line 34: I'm not crazy about this approach. Assuming I'm grok'ing correctly, this assumes that if a file exists in the `omniperf_soc` directory that follows a particular naming scheme, then it will automatically be added as a "supported" arch.  That doesn't necessarily match reality if these files are not instantiated as such in omniperf_base.py.  What if the file is a work in progress, or is a deprecated SOC in the future?

Thinking about this more, I guess our current authoritative source for supported targets is defined in omniperf_base.py/detect_soc().

Would you agree?

src/argparser.py
```diff
@@ -389,8 +410,8 @@ def omniarg_parser(parser, omniperf_home, omniperf_version):
     analyze_group.add_argument(
         "--list-metrics",
         metavar="",
-        choices=["gfx906", "gfx908", "gfx90a"],
-        help="\t\tList all available metrics for analysis on specified arch:\n\t\t   gfx906\n\t\t   gfx908\n\t\t   gfx90a",
+        choices=load_avail_arch(omniperf_home),#["gfx906", "gfx908", "gfx90a"],
+        help=print_avail_arch(load_avail_arch(omniperf_home)),
```
koomie, line 414: Thanks for this, I know I suggested we move away from the hard-coded list but not sure about the replacement based on file parsing.

src/argparser.py
coleramos425: Good point. In theory, there could be incomplete SoC definitions in `omniperf_soc`. One of the biggest considerations for an "optimal" solution is minimizing overhead for adding a new SoC. Ideally, the procedure is two steps:

1. Define an instantiation of SoC in `omniperf_soc`
2. Modify detect_soc() to consider your new SoC instance

I suppose we could add a property (a list called **__supported_archs**) to the Omniperf class such that detect_soc() is responsible for setting the value? At least this way the procedure still only requires steps (1) and (2)....

coleramos425:
Added a new property to our base class
https://github.com/AMDResearch/omniperf/blob/3d7e4bc121c2a13839f7631067fb152b817e7e28/src/omniperf_base.py#L50-L54

koomie (approved):

PR #212, opened 2023-12-13
Pytest

JoseSantosAMD:


CMakeLists.txt
```diff
@@ -245,12 +245,6 @@ install(
              src/utils/rooflines/roofline-ubuntu20_04-mi200-rocm5
     DESTINATION ${CMAKE_INSTALL_BINDIR}/utils/rooflines)
 
-# src/soc_params
```
coleramos425, line 248: Looks good. Thanks, this appears to be why the tarball test was failing

coleramos425 (approved):

PR #225, opened 2024-01-16
CLI: fix mem chart and copy config from gfx942 to gfx940 and gfx941

feizheng10:
* Fix some numbers are not shown in mem-chart Exec Block.
* Copy over config from gfx942 to gfx940 and gfx941.


coleramos425 (approved):
Looks good. Approved

PR #226, opened 2024-01-16
Add new rooflines

JoseSantosAMD:


maartenarnst:
Just a question out of curiosity. Why are the rooflines provided as binaries? I would argue it could/would be of interest and of value to users/researchers to see the source code and see how these rooflines are obtained.

feizheng10:
> Just a question out of curiosity. Why are the rooflines provided as binaries? I would argue it could/would be of interest and of value to users/researchers to see the source code and see how these rooflines are obtained.

The binary is only for measuring the peak BW on  each hierarchical level. It is "debatable" open source or not:)
The arithmetic part is open source.

coleramos425:
> The binary is only for measuring the peak BW on each hierarchical level. It is "debatable" open source or not:)
> The arithmetic part is open source.

That's right. The source code for the roofline binary reveals microarchitectural details that our hardware team has not signed off on. @maartenarnst we'll be sure to update you if that changes in the future. As Fei mentioned, the derivation for AI (arithmetic intensity) calculation is open-source if that's of any value:

https://github.com/AMDResearch/omniperf/blob/b9e474ec5ff6afcb3ebafc653dcb475acd7096d0/src/omniperf_analyze/utils/roofline_calc.py#L188

coleramos425 (approved):
Looks good to me. Approved.

PR #227, opened 2024-01-16
Upgrade plotext to 5.3.1

feizheng10:
The submodule of plotext ties to official repo now. When there is a new tag for 5.3.1, we should switch to the tag as well(or when the new package is available in pip or conda installation ).


coleramos425:
FYI @feizheng10, they _think_ timeline for PyPi module is March. We can follow up.

https://github.com/piccolomo/plotext/issues/196#issuecomment-1890988471

PR #228, opened 2024-01-17
Update list of metrics with very low variance

JoseSantosAMD:


tests/test_profile_general.py
```diff
@@ -102,63 +103,214 @@
...
-    # "7.2.3",
-    "7.2.4",
-    "7.2.7",
-    # "14.1.0",
-    "16.2.0",
-    # "16.3.14",
-    # "16.3.15",
-    "17.1.0",
+# Must not change relative difference is zero
+FIXED_METRICS = [
```
koomie, line 107: What if we tweak this data structure to define whether we want it to use an absolute tolerance or relative tolerance for pass/fail?  Would prefer to start with coarser tolerances while we gain confidence across multiple SoCs and execute across multiple nodes.

tests/test_profile_general.py
```diff
@@ -406,14 +593,34 @@ def test_path():
...
+            "16.2.0": {"absolute": 0, "relative": 7},
+            "16.3.4": {"absolute": 0, "relative": 5},
+            "17.1.2": {"absolute": 0, "relative": 3},
+            "17.2.3": {"absolute": 1, "relative": 0},
+            "17.2.14": {"absolute": 5, "relative": 0},
+            "17.3.14": {"absolute": 1, "relative": 0},
+            "17.3.15": {"absolute": 1, "relative": 0},
+            "17.3.16": {"absolute": 2, "relative": 0},
+            "18.1.13": {"absolute": 1, "relative": 0},
+            "18.1.16": {"absolute": 1, "relative": 0},
```
koomie, line 621: I think it would be useful to try and minimize how much metric/tolerance override is needed when calling the updated `logging` function.  Ideally, we would setup the global list of metrics to test once with tolerances that are mostly usable across multiple tests and override more sparingly.  Then, most checks would hopefully be shortened to look more like:

```
  logging(
        inspect.stack()[0][3],
        workload_dir,
        file_dict,
        override=None
}
```

koomie (approved):
Looks better, thx for the update.

koomie:
@JoseSantosAMD  and @coleramos425 

In order to land this PR cleanly I needed to rebase against our current 2.x branch.  During that process there were 3 conflicts.  All there of them were not related to changes to the profiling CI tests so  I deferred to the code in the latest 2.x branch.  If the intent was to make updates in the 3 files below with this PR, I apologize and would request we submit those again separately.

1. src/roofline.py
2. src/omniperf_profile/profiler_base.py
3. src/roofline.py

PR #234, opened 2024-01-30
profile: fix parsing tcc per channel on mi300 

feizheng10:
When profiling on mi300, some of the hw counter names might contain numbers as well.
The original regex doesn't work properly in that case.


coleramos425 (approved):
Looks good to me. Thanks.

PR #236, opened 2024-01-30
Improving 2.x logging

coleramos425:
Per our conversation last week, I revisiting all logging in Omniperf to move towards a consistent approach. The changes in this branch introduce dedicated logging helpers for each supported log level:

- [console_log()](https://github.com/AMDResearch/omniperf/blob/78b58f8df66db65f5d1377772ca83cb2d443e6bc/src/utils/utils.py#L57-L61)
- [console_error()](https://github.com/AMDResearch/omniperf/blob/78b58f8df66db65f5d1377772ca83cb2d443e6bc/src/utils/utils.py#L50-L56)
- [console_debug()](https://github.com/AMDResearch/omniperf/blob/78b58f8df66db65f5d1377772ca83cb2d443e6bc/src/utils/utils.py#L63-L67)
- [console_warning()](https://github.com/AMDResearch/omniperf/blob/78b58f8df66db65f5d1377772ca83cb2d443e6bc/src/utils/utils.py#L69-L70)

Now, we only import the `logging` module in src/utils/utils.py (where these helpers are defined) and utilize the helpers everywhere else. 

In our logging, sometimes we want to call out the specific module (i.e. profiling, analysis, database) that a log message is coming from for context, i.e.
https://github.com/AMDResearch/omniperf/blob/78b58f8df66db65f5d1377772ca83cb2d443e6bc/src/utils/roofline_calc.py#L147-L153
other times the context isn't needed
https://github.com/AMDResearch/omniperf/blob/78b58f8df66db65f5d1377772ca83cb2d443e6bc/src/omniperf_base.py#L109

for this, we can provide two arguments, where the first is the module to "pretty-print" in the log.

The PR also adds color to our console output for improved readability.


feizheng10:
try something like loguru?:)

PR #238, opened 2024-02-01
Adding docs updates into 2.x

coleramos425:
This PR manually stacks the changes from the docs fork (now living in `dev`) into `2.x`. After this initial merge is complete, some outstanding changes will need to be made to the branch:

1. Extend changes to perfmon configs and metrics into newly added archs in `2.x` (i.e. Mi300 SoCs)
2. Manually add typo correction commits to docs that are being overridden 
image
**Note:** Part of (2) will include modifying all references to vcopy utility in docs to reflect its new usage)
https://github.com/AMDResearch/omniperf/blob/4174788e5954a8fc96f6e6b1ffb6128f9c2fc6d5/sample/vcopy.cpp#L26-L36


coleramos425:
@skyreflectedinmirrors phase one of the merge is complete. It was straightforward other than some merge conflicts that I had to resolve in the L2 Cache per channel configs (since we're adopting a new methodology for this panel in `2.x`).

Could you give that file a quick sanity check:
https://github.com/AMDResearch/omniperf/blob/2.x-docs/src/omniperf_soc/analysis_configs/gfx90a/1800_L2_cache_per_channel.yaml

Once you confirm it looks as expected, I can continue by carrying over all the metric changes you introduced into our new Mi300 config files

coleramos425:
> 1. We're purposefully not doing the aggregates here just for review, right? I.e., making sure they look right then go on to port the agg stats?

Yes. The aggregate stats need adjusting and that will happen after this PR. As Fei reported, the std. dev. macro is incorrectly reporting (cc: @feizheng10)



> 2\. The `"::_1": 32` bit is saying "substitute this over all 32 channels", correct?

Correct.

feizheng10:
We might want to generalize it like gfx942: replace 32 to $totalL2Banks.

PR #239, opened 2024-02-05
Fix for rocm-smi return code checking

koomie:


coleramos425:
@koomie while testing this PR I found that different versions of ROCm have different return codes for `rocm-smi`. The issue is ROCm-5.2 will return 2

image

An omniperf profiling run will incorrectly break for this case

https://github.com/AMDResearch/omniperf/blob/39cce6b42b33b0a0dc1673fc884366e0a2bfcdc2/src/utils/specs.py#L232-L238

It's a pain in the butt, but we should probably add an extra conditional for this case.

koomie:
Wow, I'm at a bit of a loss for words.

PR #240, opened 2024-02-07
Install omniperf_soc and omniperf_profile, such that a 'make install'…

skyreflectedinmirrors:
…d build will function


coleramos425 (approved):
Looks good to me

PR #241, opened 2024-02-07
Escape sbios, can have commas on some systems

skyreflectedinmirrors:
Not sure if the other fields can do this either (at a guess, we might just want to escape anything that's not known to be a number), but this fixes the system-info table on my machine


feizheng10:
right, other fields are not complicated like sbios string.

coleramos425:
@skyreflectedinmirrors if you don't mind, I moved str casting into the parent function for consistency. Other than that, looks good. Thanks!

coleramos425 (approved):

PR #242, opened 2024-02-08
Workload Characterization Scripts

dwchang79:
Adding three new workload characterization python scripts:
1) workload-characterization.py - a script that parses omnitrace and omniperf data and analyzes the workload for performance bottlenecks.
2) plot-roofline.py - generates a HTML interactable roofline plot of the workload.
3) plot-characterization.py - generates two plots showing the CPU/GPU end-to-end timing and another showing the GPU bottlenecks.

Documentation can be found here: https://github.com/dwchang79/omniperf/blob/2.x/src/docs/Workload_Characterization_Manual.pdf


coleramos425:
Changing PR destination to `bottleneck-id` branch for testing and development.

PR #252, opened 2024-02-13
Fix smi runner for earlier ROCm versions

coleramos425:
Fixes issue in #239. Earlier versions of ROCm-5 (5.2) return 2 on a successful `rocm-smi`


skyreflectedinmirrors:
Solved via https://github.com/AMDResearch/omniperf/pull/273

koomie:
Believe we agreed to not worry about oddities of rocm-smi earlier than 5.7.1.  This is the minimum starting requirement for 2.x so prefer to not hack around return codes for older versions.

PR #255, opened 2024-02-14
Fix for MI300 clock timings using GRBM_GUI_ACTIVE

skyreflectedinmirrors:
Solves https://github.com/AMDResearch/omniperf/issues/248

Currently, on MI300, GRBM counters are being summed over XCDs
	- This results in many utilizations, etc., which rely on these being clock timers being incorrectly calculated (because you have ~ # XCDs * clocks)

We correct this issue by:
	- implementing a new function that maps the compute partition + arch to an XCD number based on the whitepaper
	- defining GRBM_GUI_ACTIVE_PER_XCD = average of GRBM_GUI_ACTIVE over all XCDs (and similar variables for other related counters)

For consistency, we use this in all the yaml files regardless of the arch, but on any non-MI300 arch, we should have only 1XCD, i.e., they're identical

When rocprof corrects this behavior, we can simply omit the division.

### DIFF
--- a/src/utils/specs.py
+++ b/src/utils/specs.py
@@ -69,6 +69,7 @@ class MachineSpecs:
     hbmBW: str
     compute_partition: str
     memory_partition: str
+    XCDs: str
 
     def __str__(self):
         return dedent(
@@ -105,6 +106,7 @@ class MachineSpecs:
             hbmBW:              {self.hbmBW} MB/s
             compute_partition:  {self.compute_partition}
             memory_partition:   {self.memory_partition}
+            XCCs:               {self.XCCs}
         """
         )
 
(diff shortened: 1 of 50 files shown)


src/utils/specs.py
```diff
@@ -105,6 +106,7 @@ def __str__(self):
             hbmBW:              {self.hbmBW} MB/s
             compute_partition:  {self.compute_partition}
             memory_partition:   {self.memory_partition}
+            XCCs:               {self.XCCs}
```
coleramos425, line 109:
```suggestion
            XCCs:               {self.XCDs}
```
Earlier (on line 72) you define this member var as `XCDs`. Is this a typo?

coleramos425 (changes requested):

src/utils/specs.py
skyreflectedinmirrors: Ah, yeah -- typo

skyreflectedinmirrors:
Closed via https://github.com/AMDResearch/omniperf/pull/273

PR #251, opened 2024-02-12
Various mi300 metric fixes

skyreflectedinmirrors:
Most of the fixes for MI300:

- fixes L1/L2/fabric bandwidths
- work around VMEM/FLAT instruction changes while work internally is pending
- use totalL2Banks everywhere for consistency & correctness

PR #247, opened 2024-02-12
Fix numSQCs calc

skyreflectedinmirrors:

PR #273, opened 2024-02-27
Refactor Specs + Mi300 Enhancements

coleramos425:
This PR refactors the specs.py module by cleaning up code, reducing overhead, and improving extensibility. 

- SoC-specific specs properties that require hardcoding have been relegated to their respective SoC classes
- MachineSpecs() class __init__ is now responsible for defining all class properties. Class has handy helper functions which eliminate the need for duplicate code.

Out of convenience this PR also layers on #247, #251, #252, and #255 (thanks @skyreflectedinmirrors).

2/27 TODO:

- [x] Still need an expanded explanation of MachineSpecs() properties. I'm thinking [this](https://github.com/AMDResearch/omniperf/blob/2d92bcff2e3de9e15ad2e0abf306b5d350143ae8/src/utils/specs.py#L195-L233) might be a good place to add that (since its easily accessible via `--specs`)

- [x] Still need to add some sort of version entry to MachineSpecs() output (and subsequently sysinfo.csv) to distinguish different versions of output/headers. Ideally throwing a warning when versions don't align.

- [x] I'd like to remove [this hardcoded reference](https://github.com/AMDResearch/omniperf/blob/2d92bcff2e3de9e15ad2e0abf306b5d350143ae8/src/utils/utils.py#L304-L305) to gfx90a which adds roofline to ip_blocks in sysinfo. This is in the spirit of us moving all this SoC material to subclasses. Not required, but nice to have


coleramos425:
@skyreflectedinmirrors I made some progress on L2 Cache per channel panel and the GUI filtering is now fixed. You should be good to resume verification whenever you're ready.

skyreflectedinmirrors:
Opened another PR (https://github.com/AMDResearch/omniperf/pull/280) with a few touchups, will have the specs description done today.

skyreflectedinmirrors:
RE: 

>I'd like to remove [this hardcoded reference](https://github.com/AMDResearch/omniperf/blob/2d92bcff2e3de9e15ad2e0abf306b5d350143ae8/src/utils/utils.py#L304-L305) to gfx90a which adds roofline to ip_blocks in sysinfo. This is in the spirit of us moving all this SoC material to subclasses. Not required, but nice to have

I think the easiest way is to put a flag on whether rooflines are allowed in each SOC class (if it doesn't exist already), then pass the SOC in from the profiler: https://github.com/AMDResearch/omniperf/blob/2d92bcff2e3de9e15ad2e0abf306b5d350143ae8/src/omniperf_profile/profiler_base.py#L352

We don't technically have access to the SOC from the roofline callsite (https://github.com/AMDResearch/omniperf/blob/2d92bcff2e3de9e15ad2e0abf306b5d350143ae8/src/roofline.py#L343), but I'd _assume_ roofonly is true here?

coleramos425:
> I think the easiest way is to put a flag on whether rooflines are allowed in each SOC class (if it doesn't exist already), then pass the SOC in from the profiler:

Done. Thanks for the suggestion, Nick :)

PR #287, opened 2024-02-29
Fix off by one out of range kernel filtering

JoseSantosAMD:
Related tests have also been updated

### DIFF
--- a/src/utils/parser.py
+++ b/src/utils/parser.py
@@ -852,7 +852,7 @@ def apply_filters(workload, dir, is_gui, debug):
                 if kernel_id >= len(kernels_df["Kernel_Name"]):
                     error(
                         "{} is an invalid kernel id. Please enter an id between 0-{}".format(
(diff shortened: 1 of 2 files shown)


coleramos425:
I need some additional clarification on this one. Like rocprof, Omniperf also uses zero-based indexing for its reported kernels and dispatches. i.e. see our "Top Stats" section
image

Thus, if I'm working with an app like vcopy (which only has a single kernel) then the following error makes sense. It also lines up with what's reported in "Top Stats"
image

I agree that passing 1 for the top kernel may be more logical, but I think it will cause less overall confusion if we stick with zero-based indexing everywhere.

coleramos425:
> [12:33 PM] Santos, Jose
> It would check up to the length of the kernel list
> So I would pass one and it would crash later
> Filtering isn’t off by one, the report out of range for it was

Nvm, thanks for the clarification Jose

PR #299, opened 2024-03-04
Force encoding before Omniperf execution

### LINKED ISSUE #19: investigate codec failure
Ran into this error during analyze example running  on an older ubuntu 18.04 system system that had `LANG=en_US` by default.  

```
--------
Analyze
--------

Created a Saved Analysis folder

--------------------------------------------------------------------------------
0. Top Stat
  File "/global/scratch/sw/omniperf/1.0.3/bin/omniperf", line 624, in <module>
    main()
  File "/global/scratch/sw/omniperf/1.0.3/bin/omniperf", line 604, in main
    omniperf_cli(args)
  File "/global/scratch/sw/omniperf/1.0.3/bin/omniperf_cli/omniperf_cli.py", line 225, in omniperf_cli
    tty.show_all(
  File "/global/scratch/sw/omniperf/1.0.3/bin/omniperf_cli/utils/tty.py", line 172, in show_all
    print(ss, file=output)
UnicodeEncodeError: 'latin-1' codec can't encode characters in position 0-107: ordinal not in range(256)
```

Updating to `LANG=en_US.UTF-8` fixed the issue.

We presumably always want to use utf-8 coding....

coleramos425:
Closes #19.

I read some warnings about the module not being thread-safe.

> This sets the locale for all categories to the user’s default setting (typically specified in the LANG environment variable). If the locale is not changed thereafter, using multithreading should not cause problems.

Since we always want this encoding I think it should be okay.

### DIFF
--- a/src/omniperf
+++ b/src/omniperf
@@ -25,13 +25,19 @@
 ##############################################################################el
 
 import sys
+import locale
+import logging
+from utils.utils import error
 from omniperf_base import Omniperf
 
 def main():
+    try:
+        locale.setlocale(locale.LC_ALL, 'en_US.UTF-8')
+    except locale.Error as e:
+        logging.warning("Please ensure that the 'en_US.UTF-8' locale is available on your system.")
+        error(f"Error setting locale: {e}")
 
     omniperf = Omniperf()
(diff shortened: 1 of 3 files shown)


coleramos425:
Closes #19

Issue #193, opened 2023-10-26
--kernelVerbose overwrites full kernel names in omniperf profile and analyze


IanBogle:
**Describe the bug**
The `--kernelVerbose` flag overwrites full kernel names generated by rocprof. In the profile step, this results in having to add an argument to be able to analyze with full kernel names, but `omniperf analyze` also accepts and respects `--kernelVerbose` and results in overwriting the file containing kernel names. It seems the only way to recover the longer names is to re-run `omniperf profile` with the appropriate level of --kernelVerbose.

**Development Environment:**
 - Linux Distribution: Various
 - Omniperf Version: 1.0.10, 1.1.0-PR1
 - GPU: MI200
 - Cluster (if applicable): Thera

**To Reproduce**
tar -xzvf omniperf_verbose_repro.tar.gz 
cd omniperf_verbose_repro
git clone git@github.com:kokkos/kokkos.git
cd build
./build.sh
make -j64
omniperf profile -n kokkos_dot --kernelVerbose=5 -- ./dot
omniperf analyze -p workloads/kokkos_dot/mi200 --list-kernels
(see full-length kernel names)
omniperf analyze -p workloads/kokkos_dot/mi200 --kernelVerbose=1 --list-kernels
(see shortened kernel names)
omniperf analyze -p workloads/kokkos_dot/mi200 --kernelVerbose=5 --list-kernels
(see shortened kernel names, instead of full-length kernel names)

**Expected behavior**
Since `omniperf analyze` already respects the `--kernelVerbose` argument, the shortened kernel names should be stored in a copy of the csv that holds the full kernel names. This way, `omniperf profile` will return full kernel names and shortened kernel names by default, and `omniperf analyze` can interpret the `--kernelVerbose` argument to find the file with the appropriate level of shortening (or create it if necessary). 

More generally, the full profiling data obtained from rocprof shouldn't be modified by `omniperf analyze`. Any modifications to the data should be made in such a way that the change can be reverted.

**Additional context**
[omniperf_verbose_repro.tar.gz](https://github.com/AMDResearch/omniperf/files/13182180/omniperf_verbose_repro.tar.gz)

coleramos425:
Closed in #303

PR #303, opened 2024-03-06
Kernel no overwrite

JoseSantosAMD:
This PR removes kernel-vebose from profile. It now only shows up in analyze and database mode. Instead of overwriting names in existing csvs, it reads in the csvs to a dataframe and renames kernel names in those dataframes. In database mode these updated dataframes are used instead of the csvs,. They are uploaded with pymongo to upload to mongodb


src/utils/kernel_name_shortener.py
```diff
@@ -125,18 +125,9 @@ def shorten_file(df, level):
...
-                    "profiling", "Skipping shortening on empty csv: %s" % str(fpath)
-                )
-
-        console_log("profiling", "Kernel_Name shortening complete.")
+        try:
+            modified_df = shorten_file(df, level)
+            console_log("profiling", "Kernel_Name shortening complete.")
+            return modified_df
+        except pd.errors.EmptyDataError:
+            logging.debug("[profiling] Skipping shortening on empty csv")
```
coleramos425, line 133: Instead let's use something like this with our `console_debug()` helper
```suggestion
            console_debug("profiling", "Skipping shortening on empty csv")
```

src/utils/kernel_name_shortener.py
```diff
@@ -23,18 +23,18 @@
...
 
 import os
-import glob
+import logging
 import re
 import subprocess
 import pandas as pd
 
-from utils.utils import console_error, console_debug, console_log
+from utils.utils import console_error, console_log
```
coleramos425, line 31: Keep `console_debug()`
```suggestion
from utils.utils import console_error, console_log, console_debug
```

coleramos425 (changes requested):
This PR will close an issue that a customer filed about kernel names being overridden ([#193](https://github.com/AMDResearch/omniperf/issues/193)) so could you also add a description to the PR explaining your patch and new behavior? Thanks.

JoseSantosAMD:
This PR removes kernel-vebose from profile. It now only shows up in analyze and database mode. Instead of overwriting names in existing csvs, it reads in the csvs to a dataframe and renames kernel names in those dataframes. In database mode these updated dataframes are used instead of the csvs,. They are uploaded with pymongo to upload to mongodb

coleramos425 (approved):

coleramos425:
Closes #193

PR #312, opened 2024-03-12
Check for CI_VISIBLE_DEVICES num_devices

JoseSantosAMD:
change binaries to vcopy_MI100 and vcopy_MI200


tests/vcopy_MI100
koomie, line 1: I don't think we should be adding `vcopy` binaries to the repo; what is the motivation for this?  CI steps separately build the `vcopy` executable with the correct architecture and places in the tests subdirectory.

koomie (approved):

Issue #300, opened 2024-03-05
Metric filtering does not work with baseline compare


JoseSantosAMD:
**Describe the bug**
Metric filtering does not work with baseline compare

**Development Environment:**
 - Linux Distribution:  RHEL8
 - Omniperf Version: Latest 2.x
 - GPU: MI100
 - Custer (if applicable): HPCFund

**To Reproduce**
1. Run omniperf analyze -p tests/workloads/vcopy/MI100/ -p tests/workloads/vcopy/MI200 --block 11
2.  See error numpy.core._exceptions._UFuncNoLoopError: ufunc 'add' did not contain a loop with signature matching types (dtype('float64'), dtype('<U2')) -> None
3.  Also returns error when passing same path

**Screenshots**
![image](https://github.com/AMDResearch/omniperf/assets/87447437/bc0bb59a-299c-4f03-85cf-cf888cc026ab)
![image](https://github.com/AMDResearch/omniperf/assets/87447437/d4e7c710-d7d9-4d3e-8abe-1f3e36cfce84)

**Additional context**

coleramos425:
I couldn't reproduce the issue with the commands you attached. I suspect you may have incorrectly pulled - I'll schedule an appointment to review over the phone

skyreflectedinmirrors:
I hit the same issue today on:

```
commit 62221383a701f13fac63bdbf54b8fc6fb87ebc5e (HEAD -> 2.x, origin/2.x)
Author: colramos-amd <colramos@amd.com>
Date:   Mon Mar 18 14:28:14 2024 -0500

    Fixing visualization for Compute Pipeline SOL. Need to consider 'Pct of Peak' rather than 'Avg'
    
    Signed-off-by: colramos-amd <colramos@amd.com>
```

will put details in our chat

coleramos425:
If I provide the same path twice, error is correctly caught, i.e.
```console
$ omniperf analyze -p workloads/mix/MI200/ -p workloads/mix/MI200/

  ___                  _                  __ 
 / _ \ _ __ ___  _ __ (_)_ __   ___ _ __ / _|
| | | | '_ ` _ \| '_ \| | '_ \ / _ \ '__| |_ 
| |_| | | | | | | | | | | |_) |  __/ |  |  _|
 \___/|_| |_| |_|_| |_|_| .__/ \___|_|  |_|  
                        |_|                  

Analysis mode = cli
ERROR You cannot provide the same path twice.
```
[...]

PR #331, opened 2024-03-21
Fix for #300

skyreflectedinmirrors:
On some versions of numpy/pandas, applying a str map to an empty series doesn't change the type. As a result, when adding a float to a string, sometimes numpy dies with a uadd error. We fix this by explicitly casting to a str first


coleramos425 (approved):
Verified on my end. Thanks.

PR #343, opened 2024-04-11
updates for container build with ubuntu

koomie:
* include required locale
* install python testing deps from file
* bump up rocm version

This PR also fixes the rhel-8 test to include necessary locale and correct installation order.


coleramos425 (approved):
LGTM. My only suggestion is that since now `.github/workflows/ubuntu-focal.yml` is running on Ubuntu 22.04, we may want to change the file name to `ubuntu-jammy.yml`.

Issue #344, opened 2024-04-15
Kernel menu items are too big (web GUI)


benrichard-amd:
**Describe the bug**
The height of the items in the Kernel dropdown is too big (150 pixels!). 

**To Reproduce**

omniperf analyze -p <path_to_profile> --gui

click Kernel menu drop-down

**Expected behavior**
The height of the items should be the same height as the dropdown (36 pixels).

**Screenshots**

Actual:
![menu_big](https://github.com/ROCm/omniperf/assets/143630488/50c38f4b-084d-4fa2-8f1c-b86ccc836275)

Expected (example, height = 36 pixels):
![menu_small](https://github.com/ROCm/omniperf/assets/143630488/dea6aacd-4050-4c70-a266-e44e7314ca6e)

amd-hsivasun:
Imported to ROCm/rocm-systems

PR #346, opened 2024-04-16
Fix kernel menu item size

benrichard-amd:
Addresses #344 

Change the height of the menu items from 150px to 36px.


coleramos425 (changes requested):
@benrichard-amd thanks for the contribution. While the new height does work for smaller apps with simple kernel names, we run into issues when testing apps with rocblas GEMM, Kokkos, and other complicated kernel names.

I tried loading a simple `all_reduce_perf` into the visualizer and found your branch will display overlapping text. If you'd like to change the height of the menu items, I'd recommend enabling hidden overflow to avoid this text wrapping effect.
image


[ROCm/rccl-tests](https://github.com/ROCm/rccl-tests) is a good test repo if you want to reproduce my findings.

PR #358, opened 2024-05-03
Adding a top-level runtime python dependency checker.

koomie:
Goal is to provide a kinder error message in the case where python dependencies are not available locally. This is motivated for future execution by users who are running from rocm-based binary packaging instead of using normal cmake build system which would have verified the dependencies.


coleramos425 (approved):
Looks good to me. I tried:

- Testing on a system without rocprof:
image

- Installing on a system without Python packages
image

PR #365, opened 2024-05-09
Add ability to overide arch when name missing in rocminfo

skyreflectedinmirrors:


coleramos425 (approved):
LGTM. I'll update the docs to reflect this change.

PR #375, opened 2024-06-03
Hotfixes: Adding support for new hardware

coleramos425:
This PR adds support for MI308X. Please note that `rocm-smi --showmclkrange` is currently broken for this hardware so we're hardcoding mclk until the issue is fixed. This is the same approach we take for MI100.
https://github.com/ROCm/omniperf/blob/352d1ee6993add7dfaad4be3d4cc5596cff5d228/src/omniperf_soc/soc_gfx942.py#L72-L75

To confirm backward compatibility, I've verified the changes on an mi308x node and mi300x. We can get this into mainline and the upcoming v2.0.1 release if reviewers are okay with these changes.

### DIFF
--- a/src/omniperf_soc/soc_gfx942.py
+++ b/src/omniperf_soc/soc_gfx942.py
@@ -69,6 +69,11 @@ class gfx942_soc(OmniSoC_Base):
         )
         # self.roofline_obj = Roofline(args, self._mspec)
 
+        # --showmclkrange is broken in MI308X, hardcode freq
+        if self._mspec.gpu_model == "MI308X":
+            self._mspec.max_mclk = 1300
+            self._mspec.cur_mclk = 1300
+
         # Set arch specific specs
         self._mspec._l2_banks = 16
         self._mspec.lds_banks_per_cu = 32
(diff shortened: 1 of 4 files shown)


src/omniperf_soc/soc_gfx942.py
```diff
@@ -69,6 +69,11 @@ def __init__(self, args, mspec):
         )
         # self.roofline_obj = Roofline(args, self._mspec)
 
+        # --showmclkrange is broken in MI308X, hardcode freq
```
benrichard-amd, line 72: The rocm-smi --showmclkrange fix is already in mainline. So we can exclude this work-around.

feizheng10 (approved):
LGTM

src/omniperf_soc/soc_gfx942.py
skyreflectedinmirrors: Is it in a ROCm release? More specifically: Is it in the _specific_ minumum ROCm version we expect this customer to use?

If not, we need to either hard-code or test whether it works, and use the hard-coded values as a fall-back.

skyreflectedinmirrors:
LGTM as well, at some point we should write up a "How to add an additional arch/variant" type doc.

src/omniperf_soc/soc_gfx942.py
benrichard-amd: Good point. Looks like it will be in 6.2.

We should keep this work-around in, then.

JonChesterfield:
The product number in this was tagged in a reddit thread and I can't see it under the linux tree, e.g. https://github.com/search?q=repo%3Atorvalds%2Flinux+mi308&type=code. A heads up in case it's not meant to be public.

PR #376, opened 2024-06-06
Work around crash when profiling multi-process/multi-GPU application

benrichard-amd:
Omniperf invokes `rocprof` to collect profiling data. `rocprof` has a `-o` option to specify an output CSV file. In the multi-GPU case, multiple processes will try to write to this file, corrupting it.

This change removes the `-o` option when invoking `rocprof`. Instead, Omniperf scans the output directory and combines the multiple CSV files into one CSV file.

Note: this only avoids the crash that occurs when using `rocprofv2`.

### DIFF
--- a/src/omniperf_profile/profiler_rocprof_v2.py
+++ b/src/omniperf_profile/profiler_rocprof_v2.py
@@ -48,10 +48,7 @@ class rocprof_v2_profiler(OmniProfiler_Base):
             # v2 requires output directory argument
             "-d",
             self.get_args().path + "/" + "out",
-            # v2 does not require csv extension
-            "-o",
-            fbase,
-            # v2 doen not require quotes on cmd
+            # v2 does not require quotes on cmd
             app_cmd,
         ]
         return args
--- a/src/omniperf_profile/profiler_rocprof_v2.py
+++ b/src/omniperf_profile/profiler_rocprof_v2.py
@@ -28,7 +28,6 @@ from utils.utils import (
     demarcate,
     console_log,
     replace_timestamps,
-    fixup_rocprofv2_dispatch_ids,
 )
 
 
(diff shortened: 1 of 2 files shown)


coleramos425:
@benrichard-amd in the multi-gpu case, can you confirm that we don't see any irregularities in csv output?

> Handling Indexes: By default, pd. concat preserves the original indexes of the DataFrames or Series being concatenated. This can lead to duplicate index values, which might cause issues in subsequent data operations. You can use the ignore_index=True parameter to reset the index in the resulting DataFrame.

I'm not convinced that in the multi-gpu case we'll output a different set of counters in each run, thus I'm curious how this implementation handles merging overlapping output files.

coleramos425:
Additionally, could you add a sign-off on these commits? Information on how to do so can be found under the "DCO" action attached to this PR:
image

benrichard-amd:
> @benrichard-amd in the multi-gpu case, can you confirm that we don't see any irregularities in csv output?
> 
> > Handling Indexes: By default, pd. concat preserves the original indexes of the DataFrames or Series being concatenated. This can lead to duplicate index values, which might cause issues in subsequent data operations. You can use the ignore_index=True parameter to reset the index in the resulting DataFrame.
> 
> I'm not convinced that in the multi-gpu case we'll output a different set of counters in each run, thus I'm curious how this implementation handles merging overlapping output files.

Good catch. Looking more closely we were seeing the unnamed index column in the output. This has been fixed. 

The output CSVs look good and have correct indexing.

benrichard-amd:
> Additionally, could you add a sign-off on these commits? Information on how to do so can be found under the "DCO" action attached to this PR: image

"DCO" does not appear in my check list.

image

coleramos425 (approved):
LGTM

PR #379, opened 2024-06-17
Bottleneck Characterization

coleramos425:
This PR builds off of #242 which added "Bottleneck Characterization" to Omniperf, consisting of

- End-to-end Runtime Classification plot
- GPU Kernel Runtime Breakdown plot

I have modified the code so that the characterization and plotting modules match Omniperf's class based structure. See original PR for the feature's methodology and docs

### Usage
To use the "Bottleneck Characterization" feature, just ensure omnitrace (specifically `omnitrace-instrument`) is available on your PATH. Omniperf will automatically run instrumentation as part of its profiling pipeline (disable using `--no-trace`). You can find your .proto output file in the standard Omniperf output directory.

Use analyze mode to generate the output plots (.pdf versions will also be saved by default). If for whatever reason you'd like to explicitly specify a .proto trace file you can do so in analysis using the `--trace` flag.

```console
omniperf analyze -p workloads/dummy/MI200 --trace dummy_app.proto --gui
```

You'll notice the standalone GUI view now highlights these two plots in the previously empty top section of the webpage
image

### Todo

- [x] Enable `--bottleneck-trace` option to be used without `--gui`
- [ ] Monitor known issues in https://github.com/ROCm/omniperf/discussions/378
- [x] Look for omnitrace on PATH to automate workload instrumentation 
- [ ] Update Omniperf docs for the new feature


jayhawk-commits:

vedithal-amd:
This will be tracked in https://github.com/ROCm/rocm-systems/issues/1270

PR #393, opened 2024-08-02
Check Python version on application launch

dgaliffiAMD:
On RHEL8, Omniperf fails to load after installing via the RPM package, because the default `python3` installed is version 3.6.

RPM-build changes the "shebang" from `#!/usr/bin/env python3` to `#!/usr/libexec/platform-python`, which interfered with setting virtual environments through `conda`.

Added an additional dependency check on launch. Just before we check for the PIP dependencies, we will now check the version of python being used. If it is less than 3.8, then output a more readable message, rather than a cryptic import failure that is currently seen.

Add a note to our "install via package manager" section in documentation.

Also, fixed some python linter warnings.

Internal ticket SWDEV-477233.


docs/install/core-install.rst
```diff
@@ -201,6 +201,12 @@ your operating system's native package manager using the following commands.
 See :doc:`rocm-install-on-linux:index` for guidance on installing the ROCm
 software stack.
 
+.. note::
+
+   Omniperf will use the first verion of ``Python3`` found in your system's
```
dgaliffiAMD, line 206:
```suggestion
   Omniperf will use the first version of ``Python3`` found in your system's
```

docs/install/core-install.rst
peterjunpark, line 1: lgtm

peterjunpark (approved):

docs/install/core-install.rst
```diff
@@ -201,6 +201,12 @@ your operating system's native package manager using the following commands.
 See :doc:`rocm-install-on-linux:index` for guidance on installing the ROCm
 software stack.
 
+.. note::
+
+   Omniperf will use the first version of ``Python3`` found in your system's
+   ``PATH``. If the default version of Python3 is older than 3.8, you may need to
+   update your system's ``PATH`` to point to a newer version of Python3.
```
skyreflectedinmirrors, line 208: Maybe add a note saying: If you wish to always run Omniperf with a particular version of python, we recommend (e.g.,) creating a bash alias that invokes:

```
omniperf-mypython="/path/to/my/python3 /path/to/omniperf"
```

skyreflectedinmirrors: or something like this

docs/install/core-install.rst
```diff
@@ -201,6 +201,12 @@ your operating system's native package manager using the following commands.
...
+
```
skyreflectedinmirrors, line 209: For the pip installs below, we should use:

```
python3 -m pip <...>
```

So that we ensure that the version of `pip` picked up is the same as the python3 version found in the ENV above

src/omniperf
```diff
@@ -56,15 +59,24 @@ def verify_deps_version(localVer, desiredVer, operator):
...
 def verify_deps():
-    """Utility to read library dependencies from requirements.txt and endeavor to load them within current execution environment. 
-    Used in top-level omniperf to provide error messages if necessary dependencies are not available. """
+    """Utility to read library dependencies from requirements.txt and endeavor
+    to load them within current execution environment.
+    Used in top-level omniperf to provide error messages if necessary
+    dependencies are not available."""
+
+    # Check which version of python is being used
+    if sys.version_info[0] < 3 or sys.version_info[1] < 8:
```
skyreflectedinmirrors, line 68: If there was ever a python 4, this could be problematic it's probably better to say:

```
if sys.version_info[0] < 3 or (sys.version_info[0] == 3 and sys.version_info[1] < 8):
```

skyreflectedinmirrors (changes requested):
Looks mostly good, a few minor tweaks suggested

docs/install/core-install.rst
dgaliffiAMD: That's a good idea. I think that we can move these notes to the "core installation" section. They are applicable to all installation methods.

skyreflectedinmirrors (approved):
LGTM!

PR #396, opened 2024-08-04
Fix ctest failures

benrichard-amd:
The recent improvement to counter coalescing results in Omniperf producing fewer files. This PR updates the test to refelct these changes.


coleramos425 (approved):
LGTM. Looks like you still need to format Python (e.g. `black omniperf/`) and add DCO sign-off. Once that's done, lets merge and follow up with the QA team.

PR #397, opened 2024-08-07
Disable kernel filtering for rocprofv2.

xuchen-amd:
Disable kernel filtering on omniperf profile for rocprofv2, as rocprofv2 does not currently support this feature.
Jira tracking: https://ontrack-internal.amd.com/browse/SWDEV-474758


coleramos425 (commented):
**Update:** Waiting on an update from rocprofiler team on ETA for patch. Keeping this fix in case we need to apply ahead of the next customer release.

coleramos425:
I think it's safe to close this PR now. We've confirmed the patch from rocprofiler team

PR #401, opened 2024-08-08
Promote the latest `dev` branch to `main`

dgaliffiAMD:


coleramos425:
Sorry for the late feedback, @dgaliffiAMD.

There are some-recent PRs (i.e. #400), where feature branches of feature branches did not upstream back into `dev`. Could we make sure those are merged into `dev` before syncing with `main`. In this case, we can probably cherry-pick those changes into `dev` and then sync with `main` once more?

To be fair, I don't think we've formally discussed a branching model, but moving forward this guide seems to be the gold standard for open-source projects.
[A successful git branching model](https://nvie.com/posts/a-successful-git-branching-model/)
image

PR #406, opened 2024-08-13
Check Python version on application launch

dgaliffiAMD:
Cherry pick [PR#393](https://github.com/ROCm/omniperf/pull/393)

* Check that the minimum required Python (3.8) version is used.

Prints a descriptive error message, rather than a cryptic import failure, if minimum Python version is not met.

Internal ticket SWDEV-477233.

* Disable the RPM mangling of shebangs.

The are changing the `#!/usr/bin/python3` to `#!/usr/libexec/platform-python`. With this set, omniperf is always using the platform installed version of python, which is python 3.6 on RHEL 8. Using virtual environments, like conda, did not work.

* Fix pylint issues

---------


src/omniperf
```diff
@@ -56,15 +59,24 @@ def verify_deps_version(localVer, desiredVer, operator):
...
+    to load them within current execution environment.
+    Used in top-level omniperf to provide error messages if necessary
+    dependencies are not available."""
+
+    # Check which version of python is being used
+    if sys.version_info[0] < 3 or (sys.version_info[0] == 3 and sys.version_info[1] < 8):
+        print("[ERROR] Python 3.8 or higher is required to run omniperf."
+              f" The current version is {sys.version_info[0]}.{sys.version_info[1]}.")
+        sys.exit(1)
+
```
coleramos425, line 72: Can we integrate Omniperf's built-in logging module so that we're capturing and recording this error in the same way we do elsewhere in Omniperf. i.e.

```python
from utils.utils import console_error

# Other code...

# Check which version of python is being used
    if sys.version_info[0] < 3 or (sys.version_info[0] == 3 and sys.version_info[1] < 8):
        console_error("Python 3.8 or higher is required to run omniperf."
              f" The current version is {sys.version_info[0]}.{sys.version_info[1]}.")
```

coleramos425 (changes requested):
Looks mostly good. I still have some outstanding questions on why we're updating the `release/rocm-rel-6.2` branch, but I've taken that discussion offline.

src/omniperf
coleramos425: Closer review of git blame reveals that generic handling using `print` and `sys.exit` are best suited for this function. Particularly, because this is where we handle dependency resolution; e.g. if the `logging` module is missing, we need the most generic form of error handling

coleramos425 (approved):

dgaliffiAMD:
Hi @peterjunpark , since there is an update to the docs, I'll need your review before it can me merged. Thank you. :)

peterjunpark (approved):

PR #423, opened 2024-09-10
Add the ability to determine GPU model from Chip ID

xuchen-amd:
Add the ability to determine GPU model from Chip ID to distinguish MI300 systems by using a built-in dictionary.
The built-in dictionary should be populated to include known chip id to system pairs.

### DIFF
--- a/src/omniperf_soc/soc_base.py
+++ b/src/omniperf_soc/soc_base.py
@@ -180,7 +180,7 @@ class OmniSoC_Base:
         self._mspec.gpu_model = list(SUPPORTED_ARCHS[self._mspec.gpu_arch].keys())[
             0
         ].upper()
-        if self._mspec.gpu_model == "MI300": 
+        if self._mspec.gpu_model == "MI300":
             # Use Chip ID to distinguish MI300 gpu model using the built-in dictionary
             if self._mspec.chip_id in MI300_CHIP_IDS:
                 self._mspec.chip_id = MI300_CHIP_IDS[self._mspec.chip_id]


src/omniperf_base.py
```diff
@@ -55,6 +55,11 @@
     "gfx942": {"mi300": ["MI300A_A1", "MI300X_A1"]},
 }
 
+MI300_CHIP_IDS = {
+    "29856" : "MI300A_A1",
+    "29858" : "MI308X",
+}
+
```
coleramos425: The minimum requirements for the chip id map is MI300X_A1, MI300A_A1, and MI308X. These are required because they're both indistinguishable as gfx942 arch. The others are nice to have.

Please update with MI300A_A1.

coleramos425 (changes requested):
Nice start, but I think this PR is still incomplete. To finish the PR, you'll need to update the `populate_mspec()` function which sets `gpu_model` for gfx942 archs to consider your new chip id map
https://github.com/ROCm/omniperf/blob/6bc1a3a1a1bec44d0c3fbbe3993199c071efcecc/src/omniperf_soc/soc_base.py#L187-L200

Once these changes are made, re-request my review, and I can add the original Jira ticket filer as a reviewer

src/omniperf_base.py
xuchen-amd: Hi Cole, I've added `MI300X_A1` and updated `populate_mspec()` in the latest commit. Thank you!

src/omniperf_soc/soc_base.py
```diff
@@ -197,6 +203,9 @@ def populate_mspec(self):
             # We need to distinguish MI308X by peeking reported num CUs
             elif self._mspec.cu_per_gpu == "80" or "MI308X" in self.check_arch_override():
                 self._mspec.gpu_model = "MI308X"
```
coleramos425: I don't think we need this bit anymore now that we're setting `gpu_model` based on `MI300_CHIP_IDS` map.

coleramos425 (changes requested):

coleramos425 (approved):
LGTM. Requesting a review from @skyreflectedinmirrors to see if this satisfies his original request:
https://ontrack-internal.amd.com/browse/SWDEV-461358

docs/sphinx/requirements.txt
```diff
@@ -0,0 +1,156 @@
+#
```
skyreflectedinmirrors: Why is this included in this PR? do we need it here?

readthedocs.yaml
```diff
@@ -0,0 +1,13 @@
+# Read the Docs configuration file
```
skyreflectedinmirrors: Same question: why is this part of this PR?

src/omniperf_soc/soc_base.py
```diff
@@ -194,9 +200,9 @@ def populate_mspec(self):
                 or "MI300X" in self.check_arch_override()
             ):
                 self._mspec.gpu_model = "MI300X_A1"
-            # We need to distinguish MI308X by peeking reported num CUs
```
skyreflectedinmirrors: We also need to disambiguate MI300A vs MI300X via CHIPID.  Ideally, we'd use this for all supported HW (MI200/MI100) instead of the arch

skyreflectedinmirrors (changes requested):

src/omniperf_soc/soc_base.py
xuchen-amd: In the latest commit both MI300A and MI300X are included in the chip id to gpu model dictionary. For additional support, known chip id to gpu model can be added to the dictionary for use.

readthedocs.yaml
xuchen-amd: requirements.txt and this readthedocs.yaml are added in attempt to pass a check for this PR.

src/omniperf_soc/soc_base.py
skyreflectedinmirrors: Do you have some commit locally that hasn't been pushed yet? Because I still see:

image

i.e., this is what we were trying to eliminate.  I think you could simply delete these two branches and just use the MI300_CHIP_IDS dict to set self._mspec.gpu_model?

skyreflectedinmirrors (changes requested):

readthedocs.yaml
yhuiYH: Since they didn't help pass the check, you should probably remove those files from the PR.

src/omniperf_soc/soc_base.py
xuchen-amd: Hi Nick, misunderstood the request earlier, I have removed the code that uses `gpu_arch` in determining `gpu_model`. Only the chip id dictionary is now used.

src/omniperf_soc/soc_base.py
```diff
@@ -189,25 +189,14 @@ def populate_mspec(self):
...
-                    "Cannot parse MI300 details from rocminfo. Please verify output or set the arch using (e.g.,) "
-                    'export OMNIPERF_ARCH_OVERRIDE="MI300A"'
-                )
+        # Use Chip ID to distinguish the gpu model by using built-in dictionary
+        if self._mspec.chip_id in MI300_CHIP_IDS:
+            self._mspec.chip_id = MI300_CHIP_IDS[self._mspec.chip_id]
+        else:
+            console_error(
+                "Cannot parse MI300 details from rocminfo. Please verify output or set the arch using (e.g.,) "
+                'export OMNIPERF_ARCH_OVERRIDE="MI300A"'
```
skyreflectedinmirrors: I think this env variable is no longer connected to anything , since the check_arch_override calls above were removed.  IMO, it's fine to remove it at this point, as this _should_ resolve the issue the env variable was introduced to work-around (i.e. chip-misidentification), but we may want to put that change in release notes / readme's somewhere.

src/omniperf_soc/soc_base.py
```diff
@@ -189,25 +189,14 @@ def populate_mspec(self):
             self._mspec.gpu_model = list(SUPPORTED_ARCHS[self._mspec.gpu_arch].values())[
```
skyreflectedinmirrors: I think this branch can go now.

Basically the flow is:

1. Set model from first entry in SUPPORTED_ARCHES (fine for <= MI250)
2. If MI300, set based on CHIP_ID

Not a huge problem tho, just cleaner

src/omniperf_soc/soc_base.py
skyreflectedinmirrors: No worries! Looks much better now

src/omniperf_soc/soc_base.py
xuchen-amd: Thanks Nick, cleaned up this flow and also removed the env override.

coleramos425:
Looks good. @xuchen-amd once you fix Python formatting, I will squash and merge.

PR #431, opened 2024-09-26
Reset CMakeLists.txt to correct revision and include updates from thi…

xuchen-amd:
Include python dependencies in Omniperf Debian package requirement for automatic python lib installation.


feizheng10 (approved):
LGTM.
However, as omniperf has been merged into rocm stack, if I remember correctly,  for all the packaging related, we might want to look into rocm-cmake.

PR #438, opened 2024-10-01
Automate release notes

coleramos425:
This pull request automates the creation of release notes via the [gh-release](https://github.com/marketplace/actions/gh-release) action used in the existing packaging.yml workflow. Additionally, a "smart" release name will be assigned following "Omniperf VERSION for rocm-x.x.x", i.e.
image

CC: @njobypet

### DIFF
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -57,5 +57,9 @@ jobs:
         if: startsWith(github.ref, 'refs/tags/') && github.repository == 'ROCm/omniperf'
         with:
           fail_on_unmatched: True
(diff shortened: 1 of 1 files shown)


coleramos425:
@dgaliffiAMD the `name` field introduced in this PR to assign "smart" release name could also be leveraged in Omnitraces [release.yml](https://github.com/ROCm/omnitrace/blob/1daafab8e0f6020eabf1922e35b9a0b049bd1589/.github/workflows/release.yml#L33-L40) workflow since y'all are also using softprops/action-gh-release

PR #439, opened 2024-10-02
Determine gpu model from chip ID.

xuchen-amd:
Add the ability to determine GPU model from Chip ID to distinguish MI300 systems by using a built-in dictionary.
The built-in dictionary should be populated to include known chip id to system pairs.


coleramos425 (approved):
Looks good. Approved!

PR #441, opened 2024-10-03
Skip output lines that have UTF8 decoding error 

benrichard-amd:
Omniperf assumes the output from rocprof is always UTF-8. But there is no guarantee that the user's code will always output UTF-8.

A customer was running Tensorflow and for whatever reason it output a non-UTF-8 character and crashed Omniperf.

Wrap existing code in a try/except block to catch UnicodeDecodeError so the line is skipped and profiling can continue.

### DIFF
--- a/src/utils/utils.py
+++ b/src/utils/utils.py
@@ -205,7 +205,7 @@ def capture_subprocess_output(subprocess_args, new_env=None, profileMode=False):
     buf = io.StringIO()
 
     def handle_output(stream, mask):
-        try:   
+        try:
             # Because the process' output is line buffered, there's only ever one
             # line to read when this function is called
             line = stream.readline()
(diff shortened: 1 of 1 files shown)


coleramos425 (approved):
LGTM -- just need to format Python code using `black`:
https://github.com/psf/black?tab=readme-ov-file#usage

benrichard-amd:
> LGTM -- just need to format Python code using `black`: https://github.com/psf/black?tab=readme-ov-file#usage

Done

feizheng10:
Should we force the omniperf running env with utf-8 first?

coleramos425:
> Should we force the omniperf running env with utf-8 first?

We do set project locale [here](https://github.com/ROCm/omniperf/blob/fb210abcd0133586b0b96bbb99678b6ea8491ef0/src/omniperf_base.py#L89) at Omniperf object instantiation. Is this what you're referring to?

PR #447, opened 2024-10-10
Enable baseline comparison between workloads from the same path

### LINKED ISSUE #67: Unable to compare 2 kernels from same workload
It would be nice to easily compare 2 kernels from the same workload where counters were collected for all kernels. I would like to use a command such as:
```
omniperf analyze -p workloads/vcopy_all/mi200 -k 0 -p workloads/vcopy_all/mi200 -k 1
```
This results in an error though: 
```
  File "/path/to/omniperf/dev/bin/omniperf", line 682, in <module>
    main()
  File "/path/to/omniperf/dev/bin/omniperf", line 662, in main
    analyze(args)
  File "/path/to/omniperf/dev/bin/omniperf_analyze/omniperf_analyze.py", line 253, in analyze
    run_cli(args, runs)
  File "/path/to/omniperf/dev/bin/omniperf_analyze/omniperf_analyze.py", line 195, in run_cli
    parser.load_table_data(
  File "/path/to/omniperf/dev/bin/omniperf_analyze/utils/parser.py", line 706, in load_table_data
    eval_metric(
  File "/path/to/omniperf/dev/bin/omniperf_analyze/utils/parser.py", line 570, in eval_metric
    out = eval(compile(row[expr], "<string>", "eval"))
  File "<string>", line 1
    ������@
        ^
SyntaxError: (unicode error) 'utf-8' codec can't decode byte 0x9a in position 0: invalid start byte
```
A workaround is to make a copy of this workload and use each copy in the analyze command as shown below. 
```
cp -r workloads/vcopy_all workloads/vcopy_all_2
omniperf analyze -p workloads/vcopy_all/mi200 -k 0 -p workloads/vcopy_all_2/mi200 -k 1
```
A fix would be nice to have. It is not urgent though.

zichguan-amd:
Allow same workload path to be passed twice to solve #67. Use index of workload path instead of using the workload path itself as the key to map the workloads.


feizheng10 (approved):
LGTM. Please make you verified more complicated cases like 3 ways comparison, like k0 from run 0, k1 from run 0, k2 from run 1.

coleramos425:
> Please make you verified more complicated cases like 3 ways comparison, like k0 from run 0, k1 from run 0, k2 from run 1.

@zichguan-amd please confirm you've checked the following. Once you've confirmed, we should be okay to merge. Thanks.

zichguan-amd:
> > Please make you verified more complicated cases like 3 ways comparison, like k0 from run 0, k1 from run 0, k2 from run 1.
> 
> @zichguan-amd please confirm you've checked the following. Once you've confirmed, we should be okay to merge. Thanks.

I'll update the test to include three-way comparisons

zichguan-amd:
@feizheng10 @coleramos425, I've added a new multikernel workload and three-way comparison tests

coleramos425 (approved):

PR #451, opened 2024-10-19
Add wave execution efficiency

feizheng10:
As NV has metric "wrap execution efficiency", per customer asks, we add similar metric in section Wavefront Runtime Stats.

### DIFF
--- a/src/omniperf_soc/analysis_configs/gfx906/0700_wavefront-launch.yaml
+++ b/src/omniperf_soc/analysis_configs/gfx906/0700_wavefront-launch.yaml
@@ -139,4 +139,16 @@ Panel Config:
             max: MAX((SQ_ACCUM_PREV_HIRES / $GRBM_GUI_ACTIVE_PER_XCD))
             unit: Wavefronts
             coll_level: SQ_LEVEL_WAVES
-            tips: 
+            tips:
+          Wave Execution Efficiency:
+            avg:
+              AVG(((SQ_THREAD_CYCLES_VALU / SQ_ACTIVE_INST_VALU / $wave_size * 100) if (SQ_ACTIVE_INST_VALU
+              != 0) else None))
+            min:
+              MIN(((SQ_THREAD_CYCLES_VALU / SQ_ACTIVE_INST_VALU / $wave_size  * 100) if (SQ_ACTIVE_INST_VALU
+              != 0) else None))
+            max:
+              MAX(((SQ_THREAD_CYCLES_VALU / SQ_ACTIVE_INST_VALU / $wave_size  * 100) if (SQ_ACTIVE_INST_VALU
+              != 0) else None))
+            unit: pct
+            tips: # VALU Active Threads / wavefront size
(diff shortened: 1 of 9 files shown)


src/utils/roofline_calc.py
```diff
@@ -195,11 +195,21 @@ def calc_ai(sort_type, ret_df):
     df = df.sort_values(by=["Kernel_Name"])
     df = df.reset_index(drop=True)
 
-    total_flops = valu_flops = mfma_flops_bf16 = mfma_flops_f16 = mfma_iops_i8 = (
+    total_flops = (
```
skyreflectedinmirrors, line 198: You probably don't need all these whitespace / formatting changes in the same commit?
Unless this is this linter throwing issues?

src/omniperf_soc/analysis_configs/gfx906/0700_wavefront-launch.yaml
```diff
@@ -139,4 +139,16 @@ Panel Config:
             max: MAX((SQ_ACCUM_PREV_HIRES / $GRBM_GUI_ACTIVE_PER_XCD))
             unit: Wavefronts
             coll_level: SQ_LEVEL_WAVES
-            tips: 
+            tips:
+          Wave Execution Efficiency:
```
skyreflectedinmirrors, line 143: A couple of thoughts here:

1. If we're going to add a metric, we need to update the docs in parallel.  In this case, we should simply note that this is the normalized form of VALU Active Threads
2. You probably want like: 100 * ((SQ_THREAD_CYCLES_VALU / SQ_ACTIVE_INST_VALU) / $wave_size), just for sanity / order-of-operations
3. It's sorta odd to add this here, and not in 1102 (pipeline stats), since that's where VALU Active Threads lives.  I do agree there's a bit of overlap between the two sections tho.  Open to thoughts.
4. Do we want this one, or VALU Active Threads in the top-level SOL?

skyreflectedinmirrors (changes requested):

src/utils/roofline_calc.py
feizheng10: Right. That was from my local yaml formatter. I should not avoid it as it is a noise for code reviewing.

src/omniperf_soc/analysis_configs/gfx906/0700_wavefront-launch.yaml
feizheng10: Sorry for replying late...
I totally agree with you on 1. and 2.
For 3, here are my original thoughts: (a) As a user, he/she might not care about where it is calculated from, but might want to find "Wave Execution Efficiency" under Wave Launch stats, (b) aligned with you, it is sort of duplicated in the original block 1102, which it is just a normalization of VALU Active Threads.

Combining 3. and 4., should we think about to add a new sub "speed of light" in 1102?

And, I followed the links from NV. You were right, this one seems ONLY in the old NV profiler.
I don't see NSight showing it as default. I still need to double check and verify the similar one in NCU.
In this case, I might want to go another way: cancel this commit, but mention it only in the documentation to let users know they can customize in this way.

PR #452, opened 2024-10-24
Escape the __args.remaining before joining together

HollowMan6:
The current way ignores the case and won't work if the arguments include a space.


coleramos425:
FYI @feizheng10. If I remember correctly this PR overlaps with your MPI development branch.

PR #454, opened 2024-10-28
MI300 Roofline Enablement in Omniperf

cfallows-amd:
requirements.txt
- running some modules complained about numpy version too new, adding extra requirement that numpy be 1.x pmc_roof_perf.txt
- adding TCC_BUBBLE_sum counter to profile soc_gfx940.py
soc_gfx941.py
soc_gfx942.py
- add workaround for max and current mclk to 1300MHz, cannot be read at this time from tool
- remove console logs reading that roofline is temporarily disabled, uncommenting blocks that check for roofline csv and run roofline post-processing roofline_calc.py
- add mi300 to supported soc
- add new calculation for hbm_data for MI300 using tcc_bubble_sum, checks if counter > 0 to use
- add to a few comments roofline-ubuntu-20_04-mi300-rocm6
- binary for the ubuntu systems to enable mi300 roofline calculations from rocm-amdgpu-bench

Note- other distros will get roofline bins to enable mi300, but need to be further tested before putting into branch.


coleramos425:
Overall, this looks good. My only comments are:

1. The workaround for MCLK fixed to 1300 MHz. Is this related to broken `--showmclkrange` ticket? If so, we should reconnect with rocm-smi team because this issue is likely affecting many other projects. Any reason the old conditional we had in place wasn't sufficient?
https://github.com/ROCm/omniperf/blob/75870cd3c7e1614af7d8a822030417fb0623ec92/src/omniperf_soc/soc_gfx942.py#L74-L76
2. Still need Python formatting via Black formatter (i.e. `black --diff --check .`)

coleramos425 (approved):

feizheng10 (approved):
LG. Please format the code. Thanks.

PR #457, opened 2024-11-01
Update VERSION to 3.0.0

dgaliffiAMD:


coleramos425 (approved):
Thanks, David! Only change missing from my perspective is date and version fields in README citation guide. 

https://github.com/ROCm/rocprofiler-compute/blob/31b4de1a386e4eb63f6002188c459444f6e21180/README.md?plain=1#L53-L57

Rest of references should update automatically

feizheng10 (approved):

coleramos425:
I see failing tests are a result of this block:
https://github.com/ROCm/rocprofiler-compute/blob/8a88e8bd90f4546ec48343e63f3b3758d6bcd9a5/src/utils/specs.py#L88-L98

The purpose of this code was to detect if the version attached to the profiling data (via `sysinfo.csv`) is older than the version of Rocprof-Compute being used (e.g. If you're using a newer version of the tool maybe there's new data that we've added and are expecting) and indicate the two are out of sync

I suspect the "baseline" PyTest data we have was profiled with an earlier version. To fix this I suggest we change `line 96` from the above snippet to `console_warning`.

dgaliffiAMD:
> I see failing tests are a result of this block:
> 
> https://github.com/ROCm/rocprofiler-compute/blob/8a88e8bd90f4546ec48343e63f3b3758d6bcd9a5/src/utils/specs.py#L88-L98
> 
> The purpose of this code was to detect if the version attached to the profiling data (via `sysinfo.csv`) is older than the version of Rocprof-Compute being used (e.g. If you're using a newer version of the tool maybe there's new data that we've added and are expecting) and indicate the two are out of sync
> 
> I suspect the "baseline" PyTest data we have was profiled with an earlier version. To fix this I suggest we change `line 96` from the above snippet to `console_warning`.

Correct. It's captured with version 2.y.z and only the major version is compared. That's why this version bump is triggering the error. 

I've changed it to a warning like you've suggested. Thanks!

PR #461, opened 2024-11-04
Fix versions in changelog heading

peterjunpark:
Change Omniperf version in changelog from `2.1.0` to `2.0.1` to reflect version string in actual distributed package. Associated ROCm release should be `6.2.1`, not `6.2.2`.

Also includes super minor updates to subheadings to conform to updated internal style guide.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,37 +2,37 @@
 
 Full documentation for ROCm Compute Profiler is available at [https://rocm.docs.amd.com/projects/rocprofiler-compute/en/latest/](https://rocm.docs.amd.com/projects/rocprofiler-compute/en/latest/).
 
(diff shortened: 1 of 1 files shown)


coleramos425 (approved):
Thanks Peter. Looks good.

FYI, with the current "auto populate changelog" functionality in rocprof-compute via `softprops/action-gh-release` action:
https://github.com/ROCm/rocprofiler-compute/blob/50755de8673e5f53e4e9481f6d05ff8ec8c115f1/.github/workflows/packaging.yml#L60

CHANGELOG headers will continue to generate as "Changes", "Optimizations", etc. I don't think it's a big issue, but if we _need_ to change it, I'd suggest opening a separate ticket to track.

PR #470, opened 2024-11-06
MI300 roofline enablement in rocprofiler-compute

cfallows-amd:
requirements.txt
- running some modules complained about numpy version too new, adding extra requirement that numpy be 1.x pmc_roof_perf.txt
- adding TCC_BUBBLE_sum counter to profile soc_gfx940.py
soc_gfx941.py
soc_gfx942.py
- remove console logs reading that roofline is temporarily disabled, uncommenting blocks that check for roofline csv and run roofline post-processing roofline_calc.py
- add mi300 to supported soc
- add new calculation for hbm_data for MI300 using tcc_bubble_sum, checks if counter > 0 to use
- add to a few comments roofline-ubuntu-20_04-mi300-rocm6
- binary for the ubuntu systems to enable mi300 roofline calculations from rocm-amdgpu-bench

Note- other distros will get roofline bins to enable mi300, but need to be further tested before putting into branch.


coleramos425 (approved):
Looks good. Thanks Carrie

PR #476, opened 2024-11-13
Improve formatting when displaying rocprof command.

xuchen-amd:


benrichard-amd:
This does not handle the rocprofv1 case. When we run with v1, the app cmd is surrounded in quotes: https://github.com/ROCm/rocprofiler-compute/blob/1e5cb38a9e89e877b5e23f0cba97fd84def1dc9a/src/rocprof_compute_profile/profiler_rocprof_v1.py#L51

So it will end up looking like, `rocprof "./foo --bar"` when we want `rocprof ./foo --bar`

feizheng10 (approved):

coleramos425 (approved):

coleramos425:
Before we merge, @xuchen-amd, can you rebase from `amd-staging`? I think this should solve your failing CI checks

PR #478, opened 2024-11-14
Modify numpy version in requirements.txt

cfallows-amd:
Removing numpy requirements of <2.
Checks are failing if version too high and no need for lower version.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ astunparse==1.6.2
 colorlover
 dash>=1.12.0
 matplotlib
-numpy<2,>=1.17.5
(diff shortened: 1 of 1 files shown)


cfallows-amd:
During roofline testing one of the systems was throwing error running omniperf due to pyarrow and numpy version incompatibility, tracing back to similar warning message during requirements install. I cannot reproduce this on a different machine (source machine the error came from is out of service now) and pyarrow documentation does not specify this dependency either.

coleramos425 (approved):

dgaliffiAMD (approved):

PR #480, opened 2024-11-19
MI300 rhel and sles roofline binaries

cfallows-amd:
Rhel and Sles distribution binaries from rocm-amdgpu-bench built and tested on respective MI300 distro systems.
Also minor changes removing hardcoded variables used during roofline calculations.


src/utils/roofline_calc.py
```diff
@@ -338,37 +336,38 @@ def calc_ai(sort_type, ret_df):
             )
             pass
         try:
-            if df.keys().str.contains("TCC_BUBBLE").sum() > 0:
-                # MI300 uses TCC_BUBBLE_sum to calculate hbm_data
+            if mspec.gpu_model == "MI200":
```
cfallows-amd, line 339: Diff is visually messy- change made is just checking if mi200 do original calculation, otherwise do the tcc_bubble_sum calculation for mi300 and above hardware. No changes made to the actual calculations, just the if else statement by checking for gpu model and not the counter.

src/utils/roofline_calc.py
```diff
@@ -306,8 +304,8 @@ def calc_ai(sort_type, ret_df):
             lds_data += (
                 (df["SQ_LDS_IDX_ACTIVE"][idx] - df["SQ_LDS_BANK_CONFLICT"][idx])
                 * 4
-                * L2_BANKS
-            )  # L2_BANKS = 32 (since assuming mi200 or mi300)
+                * (mspec.lds_banks_per_cu)
```
cfallows-amd, line 307: Same number (32) in the end but should be lds banks- changing to using the per-arch defined lds_banks_per_cu from soc file instead of hardcoded value.

feizheng10 (approved):

PR #483, opened 2024-11-20
[CTest] Pipeline failures for MI300

coleramos425:
This PR:
- Patches a deprecated function name left over from #475. This bug is causing the CI pipeline to fail.
- Propagates new "chip_id" logic (#423) to the CI pipeline to avoid MI300 specs parsing errors.
- Coordinated with DevOps team to ensure `ROCM_PATH` environment variable is set in Azure CI environment. Otherwise we couldn't detect versioning info
- Added `roofline.csv` to "expected files" for MI300 since it was enabled in #470

### DIFF
--- a/tests/test_profile_general.py
+++ b/tests/test_profile_general.py
@@ -38,7 +38,7 @@ config = {}
 config["rocprofiler-compute"] = SourceFileLoader(
     "rocprofiler-compute", "src/rocprof-compute"
 ).load_module()
-config["kernel_name_1"] = "vecCopy(double*, double*, double*, int, int) [clone .kd]"
+config["kernel_name_1"] = "vecCopy"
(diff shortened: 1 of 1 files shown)


tests/test_profile_general.py
```diff
@@ -319,7 +319,7 @@ def gpu_soc():
 
 # Set rocprofv2 as profiler if MI300
 if "MI300" in soc:
```
xuchen-amd, line 321: For the azure job running on gfx942 specifically, I think soc would return "MI308" here.

coleramos425: MI300X_A1. I confirmed via chip id

ywang103-amd (approved):

cfallows-amd (approved):

PR #486, opened 2024-11-25
Fix packaging for upgrading

dgaliffiAMD:
Specify that "rocprofiler-compute" replaces / obsoletes the "omniperf" package in the CPACK configuration.


raramakr (approved):

nunnikri (approved):
Looks good to me. Just a word of caution, currently 2 CMAKE COMPONENTS(main & tests) are used. But for CPACK these are not set. If in future if that is planned to package as 2 different package, then only main package needs to provides and conflict/obsoletes omniperf.

Issue #435, opened 2024-09-30
Roofline based on dispatch ID


etiennemlb:
I have profiled an application like so:

```shell
$ omniperf profile -n <case_name> --device=0 -- ./test_binary
```

Now I want, for a given dispatch, draw some rooflines:
```shell
$ # Calling this does not help as it'll generate rooflines only 
$ # on the first dispatched kernel. Even though I specified a non zero dispatch index.
$ omniperf profile -n <case_name> --roof-only --kernel-names --mem-level HBM --dispatch 546 -- ./test_binary
```

coleramos425:
Can you please share what ROCm version you're using? Additionally, could you also attach your `log.txt` file that was generated from this run for debugging purposes.

etiennemlb:
I'm using rocm 5.5.1, 5.7.1, 6.1.2.

And, Omniperf 2.1.0

sohaibnd:
Hi @etiennemlb, sorry for the delay. Can you try updating to ROCm 6.2.2+Omniperf 2.1.0 and check if the issue is still present? I followed the options you used above and was able to get the roofline analysis for a non-zero dispatch id. 

Commands Used:
`omniperf profile --name ipc_wl2 --roof-only --kernel-names --mem-level HBM --dispatch 5 -- ./ipc`
`omniperf analyze -p workloads/ipc_wl2/MI200/ --gui`

Screenshot of [standalone GUI analyzer](https://rocm.docs.amd.com/projects/omniperf/en/latest/how-to/analyze/standalone-gui.html):
![image](https://github.com/user-attachments/assets/eae66944-ed37-4679-bdc7-d983e603c74c)
![image](https://github.com/user-attachments/assets/9b52f4cd-8f49-4e3a-a137-51d5fc104701)

etiennemlb:
In the profile command above, you explicitly tell omniperf to record only for --dispatch=5. 

This is different from me recording without --dispatch specified (and thus having on disk the roofline data of all the kernels) and later trying to draw a roofline for a specific kernel. Also I can't use graphana.

sohaibnd:
I'm not sure I understand what you mean. Are you not using these options in profile mode as well?

Also, how are you accessing the roofline plot? Are you using the generated pdf directly?

etiennemlb:
I'm using the pdf directly. In fact, using graphana on a super computer tend to be cumbersome.

TLDR; I want to recod a bunch of data, and then, later, analyze it or produce roofline out of it.

[...]

PR #490, opened 2024-11-25
Move roofline plot generation to analyze mode

sohaibnd:
This is to address https://github.com/ROCm/rocprofiler-compute/issues/435

> Still, the only thing I need would be a way to generate the roofline PDFs for a given kernel after having recorded performance counters for all kernel.
> 
> I believe that would be a useful feature because we don't want to re-profile the whole program each time we want a roofline for a different kernel. This feature seems to already exist in the GUI, I would appreciate having the same capabilities built in the CLI

By moving the creation of the roofline plot pdfs to analyze mode on cli, we can create a roofline plots with different kernel filtering without having to re-profile the whole program.


coleramos425:
@sohaibnd make sure you're also updating any CI tests that leverage the flags that are modified in this PR, i.e.,
https://github.com/ROCm/rocprofiler-compute/blob/c90b6a4b8e4149f9b213d5734dc94fdbb843372b/tests/test_profile_general.py#L1575

If this introduces a new "roofline" mode, we may want to add logic for this in tests to ensure it isn't being missed

etiennemlb:
Hello, any new on that feature, it would be very useful to our use cases

jayhawk-commits:

vedithal-amd:
This will be tracked as an issue in new repo https://github.com/ROCm/rocm-systems/issues/1269

PR #492, opened 2024-11-27
part 1 to support rocprofv3

ywang103-amd:
This branch is Ben's branch that supports rocprofv3. We need to merge it into 6.4 release. I have resolved merge conflicts and pushed to my fork as branch "rocprofv3-ben".


src/rocprof_compute_profile/profiler_rocprof_v3.py
```diff
@@ -0,0 +1,92 @@
+##############################################################################bl
+# MIT License
+#
+# Copyright (c) 2021 - 2024 Advanced Micro Devices, Inc. All Rights Reserved.
```
feizheng10: Copyright (c) 2024 - 2024

src/rocprof_compute_soc/soc_base.py
```diff
@@ -393,26 +416,8 @@ def perfmon_coalesce(pmc_files_list, perfmon_config, workload_dir):
 
         pmc = []
         for block_name in f.blocks.keys():
-            if block_name == "TCC":
```
feizheng10, line 396: Does this apply to v2 still?

feizheng10 (approved):

src/rocprof_compute_soc/soc_base.py
ywang103-amd: I believe so. The unit tests are all running with v2 and they work fine

PR #499, opened 2024-12-04
fix for crash of timestamp of part 1 for rocprofv3

ywang103-amd:
for support of csv output files from rocprofv3, counter csv file is used. But when a input file does not have any counters, like "timestamp", counter csv file is not generated and this cause crash. 

A handle of this case is implemented by this branch and resolves this crash.


feizheng10 (approved):
LG

PR #508, opened 2024-12-12
Simple fix for gpu model value.

xuchen-amd:


skyreflectedinmirrors:
@feizheng10 ... this might fix the issue we were just debugging :P

I think you still want a more robust solution tho

feizheng10 (approved):

feizheng10:
> @feizheng10 ... this might fix the issue we were just debugging :P
> 
> I think you still want a more robust solution tho

@skyreflectedinmirrors  @gsitaram, I had discussion with @xuchen-amd  today: I'd suggest to merge this quick fix first.
We will provide better solution, hopefully within next week.

PR #515, opened 2024-12-17
Use Docker to setup development environment

vedithal-amd:


feizheng10 (changes requested):
General speaking, one simple principle we might want to follow is: keep the individual files as less as possible. If the Dockerfile and docker-compose.yml file are only for "dev and quick testing", I'd suggest to move those to a sub_dir in tests(even though we don't have proper structure there yet). Think about we might provide official Dockerfile for the customer to build in the future, I don't want to confuse them later.
And this Dockerfile uses ubuntu-22.04 as base line, we should mention it in the readme.

README.md
```diff
@@ -34,6 +34,10 @@ from the `amd-mainline` branch, while new features are developed in our
 
 Users may checkout `amd-staging` to preview upcoming features.
 
+## Testing
```
feizheng10, line 37: More precisely, this is more about "set up a dev test env", a developer need more info with all the cmake build steps to actually build and test.

vedithal-amd:
Thanks for the comment, Fei.

I was thinking we should move the docker files to a subdir under utils instead of tests since these are technically utility files for the project, what do you think?

feizheng10 (approved):
LGTM

README.md
vedithal-amd: Added additional instructions

PR #517, opened 2024-12-18
Auto sync staging and mainline on a weekly cadence

coleramos425:
Shameless rip-off of @dgaliffiAMD 's action in **rocprofiler-systems**. I think this is a good idea to ensure that we're keeping development in sync and don't overlook features ahead of new releases.


dgaliffiAMD:
Nothing shameless about it. :) Just note that it will fail if the "flash-forward" fails and then manual intervention is required to resolve those merge conflicts.

PR #519, opened 2024-12-18
Template for PRs

coleramos425:
With the increase in project contributors, I'd like to introduce a PR template. I want to keep it as simple as possible but also ensure everyone is on the same page as far as minimum expectations for details.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
+++ b/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
@@ -1,0 +1,9 @@
+## Describe your changes
+
+## Issue refernce (**Jira or GitHub** link)
+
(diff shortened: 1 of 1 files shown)


feizheng10:
Though we don't have a perfect solution yet...internal jira ticket info should never be exposed for now.

coleramos425:
Canceling this PR for now. Let's keep an eye on this and brainstorm future ways to keep development in sync. One note is that AMD is adopting Jira Cloud (over Jira Server) this year which will enable us to sync development across the two platforms

https://support.atlassian.com/jira-cloud-administration/docs/integrate-jira-software-with-github/

Richard, PM from rocprofiler, is tracking this but we should make sure to follow up.

PR #522, opened 2024-12-19
Pass accumulate counters to rocprofv3 using -E option

benrichard-amd:
rocprofv3 has a new -E option where extra counters can be passed instead of defining them in counter_defs.yaml.


benrichard-amd:
Don't merge -- the new rocprofv3 with the -E option is not in mainline yet.

feizheng10 (approved):

PR #537, opened 2025-01-23
Updates for new default branch

coleramos425:
This PR makes some updates to the repo (i.e. contributing, actions, readme, etc.) to reflect the new default `develop` branch. Notable changes:

- Modified [daily-staging-sync.yml](https://github.com/coleramos425/rocprofiler-compute/blob/default-branch-update/.github/workflows/daily-staging-sync.yml) action automatically merges `develop` -> `amd-staging` every night at 12AM. It will fail if it doesn't pass the --ff-only condition
- Changes to readme, contributing, and maintainers guide to reflect the new "Omnitools workflow"


vedithal-amd:
I have few concerns:

* When we push to develop branch the github workflow only tests liniting and cosmetic aspects, it does not run ctest or pytest. If this is the case, should we manually merge to staging after running the test cases to avoid sending broken code to staging.

* I think nightly auto promotion is too frequent since develop branch can see no new commits for multiple days or weeks in a row until a developer has finished a feature. A weekly or a biweekly promotion to QA (staging) would be better.

* The current auto promotion script creates a new branch and a new PR everytime the workflow runs. I think a better approach would be to just using the develop branch instead of forking a new branch and check if there is already a PR to staging before creating a PR. This would reduce the number of duplicate PRs and prevent unwanted branches.

coleramos425:
> When we push to develop branch the github workflow only tests liniting and cosmetic aspects, it does not run ctest or pytest. If this is the case, should we manually merge to staging after running the test cases to avoid sending broken code to staging.

In addition to linting, this PR also adds `develop` as a trigger to our _internal_ CTest workflows 

image

It's important we distinguish between _internal_ (Omnitool CI) and _external_ (CQE Azure) CI. Per the Omnitools workflow we discussed, the `develop` branch should trigger internal CI and `amd-staging` should trigger external CI. Currently, we do not have access to AMD hardware for internal CI, so I agree there is a gap in our coverage here. This PR respects the proposed Omnitools workflow separating the two CI types, and I think your concern related to gaps in internal CI coverage is outside the scope of this PR.

> I think nightly auto promotion is too frequent since develop branch can see no new commits for multiple days or weeks in a row until a developer has finished a feature. A weekly or a biweekly promotion to QA (staging) would be better.

Sure. I have changed it to run biweekly (Wed. 12AM and Sun. 12AM).

> The current auto promotion script creates a new branch and a new PR every time the workflow runs. I think a better approach would be to just using the develop branch instead of forking a new branch and check if there is already a PR to staging before creating a PR. This would reduce the number of duplicate PRs and prevent unwanted branches.

The new Omnitools workflow requires all contributions (w/ exception of major updates) internal+external come from forks to keep the repo clean (CC: @feizheng10). If this is the model we're using it makes sense that we test all contributions enter through develop and if approved, will be auto synced to amd-staging

vedithal-amd:
> The new Omnitools workflow requires all contributions (w/ exception of major updates) internal+external come from forks to keep the repo clean (CC: @feizheng10). If this is the model we're using it makes sense that we test all contributions enter through develop and if approved, will be auto synced to amd-staging

I am not sure what "keep the repo clean" means here, how will creating a PR using `develop` branch instead of `develop-*` forked branch make the repo "unclean".

Another question does this PR have to be approved and merged by somebody manually? We should automatically merge the PR since it is labelled as "automerge". Is there another github workflow for that? If not, we can create a "automerge" github workflow as follows, or better just merge it in this workflow:
```
run: |  
          PR_NUMBER=$(gh pr list --base amd-staging --label automerge --json number --jq '.[0].number')  
          if [ -n "$PR_NUMBER" ]; then  
            gh pr merge $PR_NUMBER --rebase 
```

We should also add a check for duplicate PR before creating a automerge PR, lets add this "if" condition
```
run: |  
          PR_EXISTS=$(gh pr list --base amd-staging --head promote-develop-$(date +%F) --json number --jq '.[0].number')  
          if [ -z "$PR_EXISTS" ]; then  
            gh pr create --base amd-staging --title "Promote \`develop\` to \`amd-staging\`" --fill --label "automerge"  
          else  
            echo "Pull request already exists: #$PR_EXISTS"  
          fi
```

coleramos425:
Thanks for the feedback! I like the suggestion to auto accept the PRs, which should help reduce overhead. That change has been made.

As for the other comments for development of feature branches, I'll take that conversation offline to settle w/in the team.

.github/workflows/daily-staging-sync.yml
```diff
@@ -0,0 +1,36 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          PR_EXISTS=$(gh pr list --base amd-staging --head promote-develop-$(date +%F) --json number --jq '.[0].number')
+          if [ -z "$PR_EXISTS" ]; then
+            git checkout amd-staging
+            git checkout -b promote-develop-$(date +%F)
+            git merge --ff-only origin/develop
+            git push -u origin HEAD
+            gh pr create --base amd-staging --title "Promote \`develop\` to \`amd-staging\`" --fill --label "automerge"
```
feizheng10: May I know why we want to create PR here?

vedithal-amd: The staging branch is protected and can only take commits via PR.
We could use a personal access token secret which has permission to bypass protection in this workflow and merge directly.
However, I think creating PR is good in terms of auditing since it keeps a record of what changes were pushed to staging and when.

vedithal-amd:
Thank a bunch for adding the changes, Cole!

Just a few final points before I approve:
- I think we can remove the check for duplicate PR since the branch name changes across workflow runs and it won't find a duplicate PR (I am assuming we always want to fork out a `develop-*` branch even though I don't think that's necessary, @feizheng10 @dgaliffiAMD  what do you think)
- We should also delete the branch after merge is complete to keep the repo clean, `git push origin --delete promote-develop-$(date +%F)` (Again assuming we want to do a fork)

If we simply use `develop` branch instead of forking a new one, then the above two changes aren't necessary

.github/workflows/daily-staging-sync.yml
```diff
@@ -0,0 +1,36 @@
+name: Sync Staging with Develop
+on:
+    workflow_dispatch:
+    schedule:
+      - cron: 0 0 * * *
```
vedithal-amd, line 5: Lets update it to biweekly, what do you think @feizheng10 @dgaliffiAMD ?

vedithal-amd: We should also rename the file accordingly

feizheng10:
> Thank a bunch for adding the changes, Cole!
> 
> Just a few final points before I approve:
> 
> * I think we can remove the check for duplicate PR since the branch name changes across workflow runs and it won't find a duplicate PR (I am assuming we always want to fork out a `develop-*` branch even though I don't think that's necessary, @feizheng10 @dgaliffiAMD  what do you think)
> * We should also delete the branch after merge is complete to keep the repo clean, `git push origin --delete promote-develop-$(date +%F)` (Again assuming we want to do a fork)
> 
> If we simply use `develop` branch instead of forking a new one, then the above two changes aren't necessary

In library team, we just do simple fastfowarding from develop branch directly:)

dgaliffiAMD:
> > Thank a bunch for adding the changes, Cole!
> > Just a few final points before I approve:
> > 
> > * I think we can remove the check for duplicate PR since the branch name changes across workflow runs and it won't find a duplicate PR (I am assuming we always want to fork out a `develop-*` branch even though I don't think that's necessary, @feizheng10 @dgaliffiAMD  what do you think)
> > * We should also delete the branch after merge is complete to keep the repo clean, `git push origin --delete promote-develop-$(date +%F)` (Again assuming we want to do a fork)
> > 
> > If we simply use `develop` branch instead of forking a new one, then the above two changes aren't necessary
> 
> In library team, we just do simple fastfowarding from develop branch directly:)

That does definitely simply things. Some example code, if this helps:

https://github.com/ROCm/AMDMIGraphX/blob/develop/.github/workflows/weekly_master_sync.yaml

vedithal-amd:
We should merge this sooner than later because formatting, RHEL 8 testing and Ubuntu Jammy testing workflows are not running on PRs to develop branch which means developers might not get CI failure for test failure.

The only thing left here to do is to update the `run` logic in `daily-staging-sync.yml`  to match that of `https://github.com/ROCm/AMDMIGraphX/blob/develop/.github/workflows/weekly_master_sync.yaml`

Approved after the above change is done.

vedithal-amd (approved):

.github/workflows/daily-staging-sync.yml
vedithal-amd: Per conversation with Fei nightly is fine

PR #541, opened 2025-01-24
Add ubuntu 24.04 detection for roofline bin loading

cfallows-amd:


feizheng10 (approved):

coleramos425 (commented):
Few things to consider:

1. Need to update docs to reflect this change
https://github.com/ROCm/rocprofiler-compute/blob/7602341b703ccb7948dfedaad7d70f854e9de4e9/docs/what-is-rocprof-compute.rst?plain=1#L91-L92
2. Modify CHANGELOG to reflect this major added OS support
3. Our workflows currently only cover Ubuntu Jammy. Makes sense to me that we expand this now that we're adding this support. Do we have a ticket to track that request?

PR #542, opened 2025-01-24
 Spatial multiplexing: part2, analysis(merge to develop)

ywang103-amd:


src/argparser.py
```diff
@@ -495,6 +495,16 @@ def omniarg_parser(
...
+    analyze_group.add_argument(
+        "--spatial-multiplexing",
+        type=int,
+        metavar="",
+        nargs="+",
+        dest="spatial_multiplexing",
+        required=False,
+        default=None,
+        help="\t\t\tProvide Node ID and GPU number per node.",
+    )
```
coleramos425: The docs need to be fully updated if we're introducing this new argument.

coleramos425 (changes requested):

src/argparser.py
```diff
@@ -495,6 +495,16 @@ def omniarg_parser(
...
     )
+    analyze_group.add_argument(
+        "--spatial-multiplexing",
+        type=int,
+        metavar="",
+        nargs="+",
+        dest="spatial_multiplexing",
+        required=False,
+        default=None,
+        help="\t\t\tProvide Node ID and GPU number per node.",
```
feizheng10: Why do we need users to specify Node ID and GPU number?

src/argparser.py
feizheng10: > The docs need to be fully updated if we're introducing this new argument.

For all doc tasks, it seems we need to work with Doc team together.

src/utils/utils.py
```diff
@@ -1084,3 +1084,118 @@ def set_locale_encoding():
...
+
+        # Append the DataFrame for this level
+        dfs.append(columns_for_level)
+
+    # Return the list of DataFrames and the column levels
+    return dfs, coll_levels
+
+
+def merge_counters_spatial_multiplex(df_multi_index):
+    non_counter_column_index = [
```
feizheng10, line 1126: No func comments

src/argparser.py
coleramos425: CC: @peterjunpark 

also, related to issue #548

src/utils/utils.py
ywang103-amd: added comments

src/argparser.py
ywang103-amd: I made a new commit that make this argument "store_ture" so user will not need to specify node id and gpu number anymore

src/argparser.py
ywang103-amd: let's work with doc team later on to update the documents.

feizheng10 (approved):

vedithal-amd (approved):

vedithal-amd:
Merging while bypassing approval by @coleramos425 per my conversation with @ywang103-amd since this is urgent.

PR #546, opened 2025-01-28
Standalone binary

vedithal-amd:
Feature:
* Package rocprof-compute python code, configuration/data files, roofline binaries and python dependencies into a single files using Nuitka
* This will allow using rocprof-compute functionalities using a single binary files without requiring to install python dependencies and the only requirement is glibc version >= 2.28
* libnss3.so is required to generate PDF file containing roofline data

Changes:
* Add cmake function to create standalone binary using Nuitka command line tool
* Add docker files to create the environment to build the standalone binary
* Add README instructions to build the standalone binary, also added requirements and instructions to test
* Add third party python and Nuitka dependencies to LICENSE file
* Add test cases to run automated test suite on the binary by passing --call-binary option to pytest

Checkilst:
[x] Delete vedithal/standalone-binary branch from this repo, it was created for OSRB approval and security scan


utils/docker_env/docker-compose.standalone.yml
coleramos425, line 1: I'd encourage we keep all general Docker related assets in the root level `Docker/` directory. This is where we keep misc. Docker assets related to the project (i.e. containers we use for CI env). You could introduce subdirectories to separate CI Docker assets from project binary assets

utils/docker_env/Dockerfile.standalone
coleramos425, line 1: Same comment as `utils/docker_env/docker-compose.standalone.yml`

tests/test_binary_profile_general.py
coleramos425, line 1: Duplicating each PyTest module for binary testing is a bit extreme, not to mention unwieldy for project maintainers. We can re-use existing PyTest modules by adding a conditional to check if this is a "binary" or "regular" test and call the appropriate function from your util function

tests/test_binary_analyze_workloads.py
coleramos425, line 1: same comment as other file

tests/test_binary_analyze_commands.py
coleramos425, line 1: same comment as other file

coleramos425 (changes requested):
Good that we're updating the testing accordingly, but I think there's a more elegant approach (below). 

Side note, if we're now supporting a standalone binary, we may want to consider adding this as an asset to future releases. In that case we'd need a workflow to create that binary asset and then the `packaging.yml` action would need to attach it to the release in its final step
https://github.com/ROCm/rocprofiler-compute/blob/0a52e6f19d2b4f4e81a66fb419b522d32f210abf/.github/workflows/packaging.yml#L55-L64

utils/docker_env/Dockerfile.standalone
```diff
@@ -0,0 +1,18 @@
+FROM compute-artifactory.amd.com:5000/rocm-base-images/rhel-8.x-bld:2024112501
```
feizheng10: Should we refer to the public  docker?

tests/test_binary_profile_general.py
vedithal-amd: I used pytest fixture to conditionally invoke binary or main function based on whether an optional --call-binary option is passed

utils/docker_env/Dockerfile.standalone
vedithal-amd: I switched to public docker and added the tag of the most recent version for stability

utils/docker_env/docker-compose.standalone.yml
vedithal-amd: I moved all docker related files to the appropriate folder

vedithal-amd:
> Good that we're updating the testing accordingly, but I think there's a more elegant approach (below).
> 
> Side note, if we're now supporting a standalone binary, we may want to consider adding this as an asset to future releases. In that case we'd need a workflow to create that binary asset and then the `packaging.yml` action would need to attach it to the release in its final step
> 
> https://github.com/ROCm/rocprofiler-compute/blob/0a52e6f19d2b4f4e81a66fb419b522d32f210abf/.github/workflows/packaging.yml#L55-L64

We do not plan to do general release of the standalone binary yet until we talk to other ROCm component owners and build alignment about using standalone binary. Hence, there is no need for automated artifact creation workflow. For more detail on how the standalone binary will be distributed please check the project ticket.

vedithal-amd:
@feizheng10 @coleramos425 could you please approve this PR?
I have resolved the outstanding issues, please let me know if there are any more issues.

requirements-standalonebinary.txt
coleramos425, line 1: Since requirements-standalonebinary.txt would only really ever be called in the generation of standalone binary via the Dockerfile, how would y'all feel about moving it out of the root directory and into `docker/`? Reason being, this would keep the root directory cleaner and avoid any confusion with core project dependencies.

In the same vein, if we're moving that inside `docker/`, I'd propose we also introduce two subdirectories to organize those files. Files in green live in `docker/standalone-binary/`, files in red live in `docker/build-env/`
image

coleramos425 (approved):
Looks good. Only suggestion is we add some organization the the growing `docker/` subdir. Also, do we want to make an update to CHANGELOG?

requirements-standalonebinary.txt
vedithal-amd: We should definitely advertise this feature in CHANGELOG. One thing to note, with this PR we are adding capability to generate a standalone binary using docker but not shipping the binary itself. This is a useful feature for our customers since they only need to build the binary on one system and then they can distribute the binary throughout the cluster for easy usage without environment setup.

The standalone binary can also be built without docker by running cmake commands on the local system. Docker just ensures a consistent glibc environment for binary building, but that's not necessary. All the python requirement files should be together for easy of understanding the software dependencies. If we are worried about too many requirements files in the root directory, we can move them under a single requirements folder @coleramos425?

The docker files for testing are not related to standalone binary in anyway, they were added in prior PR for easily getting a test environment for developers. So, I don't think they should be grouped together. We can create two subirs, one for testing and one for standalone binary. @coleramos425 do you know what are the existing docker files used for?

Also adding @feizheng10 for his comments on this.

coleramos425: Sounds good. I'll let y'all make a judgement call on this. Merge whenever you're ready

CMakeLists.txt
```diff
@@ -372,6 +372,27 @@ add_custom_target(
         "src/${PACKAGE_NAME},cmake/Dockerfile,cmake/rocm_install.sh,docker/docker-entrypoint.sh,src/rocprof_compute_analyze/convertor/mongodb/convert"
     )
 
+# Standalone binary creation
+add_custom_target(
+    standalonebinary
+    # Change working directory to src
+    WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/src
+    # Check nuitka
+    COMMAND ${Python3_EXECUTABLE} -m pip list | grep -i nuitka > /dev/null 2>&1
```
feizheng10, line 381: You might want to add comments: this doesn't work on WIN.

feizheng10 (approved):
Great job!

vedithal-amd:
RHEL 8 test is failing due to reason unrelated to this PR as shown below, hence bypassing this test for merging:

Errors during downloading metadata for repository 'amdgpu':
  - Status code: 404 for https://repo.radeon.com/amdgpu/latest/rhel/8.8/main/x86_64/repodata/c66aaff110b9902107ca5fa6ab662552eab7d5bf4cb04fe7d4a5ca4518f86b28-filelists.xml.gz (IP: 23.53.11.171)
  - Status code: 404 for https://repo.radeon.com/amdgpu/latest/rhel/8.8/main/x86_64/repodata/00766b3f175c4d456f6e5c08bac72716e20551df476da93c9c1858cdc9b1de50-primary.xml.gz (IP: [23](https://github.com/ROCm/rocprofiler-compute/actions/runs/13445904711/job/37571053704?pr=546#step:3:24).53.11.171)
Error: Failed to download metadata for repo 'amdgpu': Yum repo downloading error: Downloading error(s): repodata/00766b3f175c4d456f6e5c08bac72716e20551df476da93c9c1858cdc9b1de50-primary.xml.gz - Cannot download, all mirrors were already tried without success; repodata/c66aaff110b9902107ca5fa6ab662552eab7d5bf4cb04fe7d4a5ca4518f86b28-filelists.xml.gz - Cannot download, all mirrors were already tried without success
Error: Process completed with exit code 1.

PR #553, opened 2025-01-31
Use GitHub app for develop to amd-staging sync

vedithal-amd:
* Use rocprofile-compute-runner GitHub app for authentication to bypass PR code review approval for auto merge

### DIFF
--- a/.github/workflows/daily-staging-sync.yml
+++ b/.github/workflows/daily-staging-sync.yml
@@ -10,15 +10,21 @@ jobs:
     runs-on: ubuntu-latest
     name: Promote Develop to Staging
     steps:
+      - name: Generate a token
+        id: generate-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ secrets.ROCPROFILER_COMPUTE_RUNNER_APP_APP_ID }}
+          private-key: ${{ secrets.ROCPROFILER_COMPUTE_RUNNER_APP_PRIVATE_KEY }}
+
       - name: Checkout
         uses: actions/checkout@v4
         with:
           ref: develop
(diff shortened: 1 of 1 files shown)


vedithal-amd:
I did some research and seems like we dont really need a PAT, we can create a github app for CI in rocprofiler-compute, then install the github app in the repo, add the github app secrets to the repo, then update the workflow to use github app secrets for authentication. All of this is explained in [Making authenticated API requests with a GitHub App in a GitHub Actions workflow - GitHub Docs](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/making-authenticated-api-requests-with-a-github-app-in-a-github-actions-workflow#about-github-actions-authentication). I have done these steps.

One more thing, it seems like we do have to create a PR and auto merge (rebase merge) the PR for auto promotion. This is mandated by staging branch protection, they bypass option you see doesn't really bypass creating the PR in the first place, it actually bypasses approval conditions in the PR before merging it.

@feizheng10 , I know you were against creating the auto merge PR but I don't think we can avoid it.

vedithal-amd:
> One more thing, it seems like we do have to create a PR and auto merge (rebase merge) the PR for auto promotion. This is mandated by staging branch protection, they bypass option you see doesn't really bypass creating the PR in the first place, it actually bypasses approval conditions in the PR before merging it.

> @feizheng10 , I know you were against creating the auto merge PR but I don't think we can avoid it.

I was wrong in claiming above, thanks to experiments performed by @coleramos425 it seems that we need to add the token during checkout step but not during merge step

coleramos425 (approved):

feizheng10 (approved):

PR #556, opened 2025-02-07
Add broken mclk workaround for all arch #453

vedithal-amd:
* Add the missing commit from #453

### DIFF
--- a/src/rocprof_compute_soc/soc_gfx908.py
+++ b/src/rocprof_compute_soc/soc_gfx908.py
@@ -65,8 +65,10 @@ class gfx908_soc(OmniSoC_Base):
         self._mspec.lds_banks_per_cu = 32
         self._mspec.pipes_per_gpu = 4
         # --showmclkrange is broken in Mi100, hardcode freq
-        self._mspec.max_mclk = 1200
-        self._mspec.cur_mclk = 1200
+        if self._mspec.gpu_model == "MI100":
+            if self._mspec.max_mclk == None or self._mspec.cur_mclk == None:
+                self._mspec.max_mclk = 1200
+                self._mspec.cur_mclk = 1200
 
     @demarcate
     def get_profiler_options(self):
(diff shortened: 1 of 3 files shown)


feizheng10 (approved):

src/rocprof_compute_soc/soc_gfx908.py
```diff
@@ -65,8 +65,10 @@ def __init__(self, args, mspec):
         self._mspec.lds_banks_per_cu = 32
         self._mspec.pipes_per_gpu = 4
         # --showmclkrange is broken in Mi100, hardcode freq
-        self._mspec.max_mclk = 1200
-        self._mspec.cur_mclk = 1200
+        if self._mspec.gpu_model == "MI100":
```
coleramos425: Do we need this line 68 conditional? If we've instantiated gfx908 class, `gpu_model == "MI100"` will always hold.

src/rocprof_compute_soc/soc_gfx90a.py
```diff
@@ -74,6 +74,13 @@ def __init__(self, args, mspec):
         )
         self.roofline_obj = Roofline(args, self._mspec)
 
+        # Workaround for broken --showmclkrange
+        # MI210/MI250/MI250X have 1600MHz mclk
+        if self._mspec.gpu_model == "MI200":
```
coleramos425: Same comment as above. gfx90a will always be `gpu_model == "MI200"`

coleramos425 (approved):

src/rocprof_compute_soc/soc_gfx908.py
vedithal-amd: I agree that we dont need the conditional when looking at https://github.com/ROCm/rocprofiler-compute/blob/3396ba39064afe5946a9ccf37fb9130ad0bc0cfd/src/rocprof_compute_soc/soc_base.py#L186 and https://github.com/ROCm/rocprofiler-compute/blob/3396ba39064afe5946a9ccf37fb9130ad0bc0cfd/src/rocprof_compute_base.py#L58

vedithal-amd: We should also use `is` instead of `==` when comparing to `None`

vedithal-amd: See https://stackoverflow.com/questions/14247373/python-none-comparison-should-i-use-is-or/67592640

PR #557, opened 2025-02-10
fix crash for running rocprofv3 on mi100

ywang103-amd:
removed "-m" parameter for rocprofv3 when on mi100 and add mechanism to avoid crash when rocprofv3 does not generate counter csv file.

### DIFF
--- a/src/rocprof_compute_soc/soc_gfx908.py
+++ b/src/rocprof_compute_soc/soc_gfx908.py
@@ -25,7 +25,7 @@
 from pathlib import Path
 
 import config
-from rocprof_compute_soc.soc_base import OmniSoC_Base, using_v3
+from rocprof_compute_soc.soc_base import OmniSoC_Base
 from utils.utils import console_error, demarcate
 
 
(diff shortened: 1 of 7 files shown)


src/rocprof_compute_soc/soc_gfx908.py
```diff
@@ -25,7 +25,7 @@
 from pathlib import Path
 
 import config
-from rocprof_compute_soc.soc_base import OmniSoC_Base
+from rocprof_compute_soc.soc_base import OmniSoC_Base, using_v3
```
coleramos425: I don't encourage importing dirty functions from parent class like this. Let's refine this approach and reconsider.

coleramos425 (changes requested):

src/rocprof_compute_soc/soc_gfx908.py
ywang103-amd: removed in latest version

feizheng10 (approved):

coleramos425 (approved):

PR #573, opened 2025-02-20
Improve kokkos trace handling.

xuchen-amd:
- refactored `run_prof()` to improve readability for rocprofv3 output processing
- added `--kokkos-trace` output processing for rocprofv3 support 
- updated `join_prof()` with `--hip-trace` and `--kokkos-trace` compatibility
- tested locally with ctest

**to be merged together with #588, fix output processing issue**


src/rocprof_compute_profile/profiler_base.py
```diff
@@ -77,6 +78,22 @@ def join_prof(self, out=None):
                 out = self.__args.path + "/pmc_perf.csv"
             files = glob.glob(self.__args.path + "/" + "pmc_perf_*.csv")
             files.extend(glob.glob(self.__args.path + "/" + "SQ_*.csv"))
+
+            if self.get_args().hip_trace:
+                # remove hip api trace ouputs from this list
+                files = [
+                    f
+                    for f in files
+                    if not fnmatch.fnmatch(os.path.basename(f), "*_hip_api_trace.csv")
```
feizheng10: Not a big deal...why fnmatch but not regex?
If there is portable solution, we should to go for it now in case WIN support in the future:)

src/utils/utils.py
```diff
@@ -757,6 +699,89 @@ def run_prof(
...
+                )
+
+                v3_counter_csv_to_v2_csv(
+                    counter_file, agent_info_filepath, converted_csv_file
+                )
+
+            results_files_csv = glob.glob(workload_dir + "/out/pmc_1/*/*_converted.csv")
+        elif is_timestamps:
+            # when the input is timestamps, we know counter csv file is not generated and will instead parse kernel trace file
+            results_files_csv = glob.glob(
```
feizheng10, line 752: Not a big deal but path lib should be better than glob

src/utils/utils.py
```diff
@@ -757,6 +699,89 @@ def run_prof(
     df.to_csv(workload_dir + "/" + fbase + ".csv", index=False)
 
 
+def process_rocprofv3_output(rocprof_output, workload_dir, is_timestamps):
+    results_files_csv = {}
```
feizheng10, line 707: Add func comments with ''' '''

feizheng10 (commented):
Some minor questions.

src/utils/utils.py
xuchen-amd: `results_files_csv` is returned and to expected to be looped through as string type to pass to `pd.read_csv()`. In future refactoring, I can address this problem to handle paths with more robustness.

feizheng10 (approved):
Please make sure CI passed.

PR #581, opened 2025-02-25
Improve chip info logic.

xuchen-amd:
### This PR:
- Purpose: centralize and unify the source of system/chip related information.
- Added YAML file [src/utils/mi_gpu_data.yaml](https://github.com/ROCm/rocprofiler-compute/compare/develop...xuchen-amd:rocprofiler-compute:improve_chip_info_logic#diff-1436f31cbd54dd4e15316d16583553995db3115543e295c61e035113fcf39a76) to promote readability and scalability for future supports.
- Added [src/utils/mi_gpu_data.py](https://github.com/ROCm/rocprofiler-compute/compare/develop...xuchen-amd:rocprofiler-compute:improve_chip_info_logic#diff-2c0ae48d1498ea68804c3718b574751335865173b91160403aa300b27cbb3e01) for system/machine spec retrievals.
- Removed unused data structures and functions.
- Added debug messages and fixed formatting.
- Passed local ctest.
### WIP:
- [ ] Instantiate MIGPU singleton for the current system for easy references.
- [ ] Display graphic information of the system by visualize the MIGPU singleton using Dash.


src/rocprof_compute_base.py
```diff
@@ -42,6 +41,10 @@
     setup_file_handler,
     setup_logging_priority,
 )
+from utils.mi_gpu_data import (
```
feizheng10: mi_gpu_data mi_gpu_spec seems better to me.

src/utils/mi_gpu_data.py
```diff
@@ -0,0 +1,304 @@
...
+    """
+    Represents the compute partition mode.
+    """
+
+    def __init__(self, num_xcds=None):
+        self.__num_xcds = num_xcds
+
+    def get_num_xcds(self):
+        return self.__num_xcds
+
```
feizheng10: As I suggested, regardless using or not, we need to define MemoryPartitionMode as well.

src/utils/mi_gpu_data.yaml
```diff
@@ -0,0 +1,144 @@
...
+#               |-- chip_ids: chip id is specific to the environment the gpu is being used on
+#               |-- mi300_arch: mi300 specific architectures; mi300a, mi300x
+#                   |-- partition_mode: currently, only mi300 gpus hold partition mode information
+#                                       two types: compute partition mode, memory partition mode,
+#                                       currently only mi300 gpus contains compute partition mode information on number of xcds
+#
+# --------------------------------------------------------------------------------
+
+mi_gpu_data:
+  - gpu_series: "mi50"
```
feizheng10: Does the "" is mandatory? or mi50 itself is simple and good enough?

src/utils/mi_gpu_data.yaml
```diff
@@ -0,0 +1,144 @@
...
+#   |-- series: the specific MI series; mi50, mi100, mi200, mi300
+#       |-- architecture: currently, only mi300 gpus hold different architectures
+#           |-- models
+#               |-- chip_ids: chip id is specific to the environment the gpu is being used on
+#               |-- mi300_arch: mi300 specific architectures; mi300a, mi300x
+#                   |-- partition_mode: currently, only mi300 gpus hold partition mode information
+#                                       two types: compute partition mode, memory partition mode,
+#                                       currently only mi300 gpus contains compute partition mode information on number of xcds
+#
+# --------------------------------------------------------------------------------
```
feizheng10: Probably, we might want to add a comment here, something like:
This file is stack holder of  some common spec for one GPU series, not all!
For example, the CU # would be based on info retrieved from other tools.

src/utils/specs.py
```diff
@@ -37,13 +38,18 @@
 import pandas as pd
 
 import config
+from utils.mi_gpu_data import (
+    get_gpu_series_dict,
+    get_mi300_chip_id_dict,
```
feizheng10, line 43: Not sure it is better or not: general interface like get_chip_id(gpu_arch)?

feizheng10 (commented):
Cool! The whole structure is much better than before!

src/utils/mi_gpu_data.py
xuchen-amd: memory partition and chip ids for different env are preserved fields in the yaml file for scalability :)

src/utils/utils.py
```diff
@@ -1029,51 +1031,6 @@ def flatten_tcc_info_across_xcds(file, xcds, tcc_channel_per_xcd):
     return df
 
 
-def total_xcds(archname, compute_partition):
```
feizheng10, line 1054: We might want to keep total_xcds on top of the new get_mi300_num_xcds() for any other archs even they don't have the concept of "xcd". Think about future arch, people still probably want to call the unified interface from total_xcds

feizheng10 (approved):

feizheng10:
Please make sure CI passed. Thanks!

src/rocprof_compute_soc/soc_base.py
```diff
@@ -102,7 +105,8 @@ def get_compatible_profilers(self):
 
     @demarcate
     def populate_mspec(self):
-        from utils.specs import run, search, total_sqc, total_xcds
+        console_debug("[populate_mspec]")
```
coleramos425, line 108: The purpose of the `@demarcate` decorator is to help developers debug in trace mode, the call stack. Correct me if I'm wrong, but I believe this overlaps with the change you made on line 108 with the debug statement with function name.

My advice is that we remove line 108, and moving forward, any "critical" functions should use `@demarcate` decorator to facilitate call stack tracing

Issue #576, opened 2025-02-22
[Req]: Can we unify the option for workload name in profile and analyze modes?


gsitaram:
### Is your feature request related to a problem?

Could we just use the same option for workload name ? Today in `profile` mode, we have to say `-n`, and in `analyze` mode we have to say `-p`. Also, `-n` requires just a workload name, but `-p` requires a path. Is it possible to make this simpler?

### Describe the solution you'd like

I propose using `-n <workload_name>` in both cases. In `analysis` mode, this would go and find the sub-directory called `workloads/<workload_name>/<arch>` and do the analysis. Here, `<arch>` will have to be determined by rocprof-compute the same way it finds the string to use in `profile` mode without the user's help.

PR #584, opened 2025-02-26
Add name as positional argument

darren-amd:
This change aims to add the name as a positional argument rather than specifying it with -n or --name. Also adds --name for analyze mode.

Originated from https://github.com/ROCm/rocprofiler-compute/issues/576


feizheng10 (approved):
LGTM. I'd say we need mention details in documentation.

PR #589, opened 2025-03-02
Fix counter collection inconsistency with rocprofv3 

feizheng10:
For some case, we noticed collected total row numbers are not the same across all perfmon csv files:
pmc_perf.csv, SQ_IFETCH_LEVEL.csv, SQ_INST_LEVEL_LDS.csv, SQ_INST_LEVEL_SMEM.csv, SQ_INST_LEVEL_VMEM.csv.
The quick solution is: inner join if collected csv are not on the same page.
We might need to dig into and check the root reason.


src/rocprof_compute_soc/soc_base.py
```diff
@@ -304,9 +304,9 @@ def add(self, counter) -> bool:
         return self.blocks[block].add(counter)
 
 
-# TODO: This is a HACK
+# FIXME: This is a HACK
 def using_v3():
-    return "ROCPROF" in os.environ.keys() and os.environ["ROCPROF"] == "rocprofv3"
+    return "ROCPROF" in os.environ.keys() and "rocprofv3" in os.environ["ROCPROF"]
```
vedithal-amd: Just curious as to why does this hack do substring match?

ywang103-amd (approved):

src/rocprof_compute_soc/soc_base.py
vedithal-amd: I get it now, ROCPROF env. var. will contain the path to rocprofv3, in that case should we use `os.environ["ROCPROF"].endswith("rocprofv3")`

src/utils/file_io.py
```diff
@@ -209,7 +209,8 @@ def create_single_df_pmc(raw_data_dir, node_name, kernel_verbose, verbose):
                     dfs.append(tmp_df)
                     coll_levels.append(f[:-4])
 
-        final_df = pd.concat(dfs, keys=coll_levels, axis=1, copy=False)
+        # TODO: double check the case if all tmp_df.shape[0] are not on the same page
+        final_df = pd.concat(dfs, keys=coll_levels, axis=1, join="inner", copy=False)
```
vedithal-amd, line 213: As discussed in the meeting, inner join based on index is the only option we have since kernel name and dispatch ids might be non-deterministic

vedithal-amd (approved):
LGTM, just some minor comments

src/rocprof_compute_soc/soc_base.py
feizheng10: [bc9cf07](https://github.com/ROCm/rocprofiler-compute/pull/589/commits/bc9cf07dc93392364b8aa7108f1fc2769f80542c)

PR #592, opened 2025-03-04
FP8 roofline support

cfallows-amd:
Adding FP8 datatype to roofline in rocprof-compute on MI300-based systems.
FP8 listed in terminal output and roofline csv, and outputs a standalone PDF.


feizheng10 (approved):
LG. We need refactoring the code later:)

coleramos425 (approved):

PR #595, opened 2025-03-06
fix specs-correction

feizheng10:
specs-correction is an option for end user to correct any attr fro machine spec manually.
Some of the attr, like sys_info.cu_per_gpu need to be converted to number from str before using it.


vedithal-amd (approved):
LGTM

Could you also add in PR description why this change is necessary

coleramos425 (approved):

feizheng10:
> LGTM
> 
> Could you also add in PR description why this change is necessary

Sorry about that. Just add descriptions. 
It was working...

xuchen-amd (approved):

PR #597, opened 2025-03-07
TCC per channel metric's fix for rocprofv3

ywang103-amd:
implemented way of generating self-defied intermediate counter in yaml file for -E parameter in order for rocprofv3 to provide single channel's counter metric of TCC counters

### DIFF
--- a/src/utils/utils.py
+++ b/src/utils/utils.py
@@ -46,17 +46,20 @@ import config
 rocprof_cmd = ""
 rocprof_args = ""
 
+
 # TODO: This is a HACK
 def using_v3():
     return "ROCPROF" in os.environ.keys() and "rocprofv3" in os.environ["ROCPROF"]
 
+
 def get_global_yaml_config_file():
     return os.path.join(
-                str(config.rocprof_compute_home),
-                "rocprof_compute_soc",
-                "profile_configs",
-                "accum_counters.yaml",
-            )
+        str(config.rocprof_compute_home),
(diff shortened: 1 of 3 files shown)


src/utils/utils.py
```diff
@@ -47,6 +47,20 @@
 rocprof_args = ""
 
 
+# TODO: This is a HACK
+def using_v3():
+    return "ROCPROF" in os.environ.keys() and "rocprofv3" in os.environ["ROCPROF"]
+
+
+def get_global_yaml_config_file():
+    return os.path.join(
```
feizheng10: Never use os.join please

src/utils/utils.py
```diff
@@ -47,6 +47,20 @@
 rocprof_args = ""
 
 
+# TODO: This is a HACK
+def using_v3():
+    return "ROCPROF" in os.environ.keys() and "rocprofv3" in os.environ["ROCPROF"]
+
+
+def get_global_yaml_config_file():
```
feizheng10: No definition comments for this func.
And the func name is too generic.

src/utils/utils.py
ywang103-amd: changed it to Path.join_path

src/utils/utils.py
ywang103-amd: changed the function's name and added comments.

feizheng10 (approved):

PR #599, opened 2025-03-10
Force kaleido version to be no greater than 0.2.1

cfallows-amd:
Higher versions (eg. 0.4.1) have external dependencies that are causing errors and forcing early exits without creating roof plots


vedithal-amd (approved):
LGTM

PR #604, opened 2025-03-11
Force kaleido version to be no greater than 0.2.1

cfallows-amd:
Higher versions (eg. 0.4.1) have external dependencies that are causing errors and forcing early exits without creating roof plots


coleramos425 (approved):

feizheng10:
Can we do something like <=0.4.1?

feizheng10 (approved):

PR #605, opened 2025-03-12
amd-smi replacing rocm-smi in rocprof-compute

ywang103-amd:


src/rocprof_compute_soc/soc_base.py
```diff
@@ -27,6 +27,7 @@
 import os
 import re
 import shutil
+from amdsmi import *
```
feizheng10: First, we should not import  *.
Second, we might want to try and catch the exception, as a defending of dependency. If it is not there, we should log the warning and allow user to keep profiling.

src/rocprof_compute_soc/soc_base.py
```diff
@@ -164,8 +165,24 @@ def populate_mspec(self):
         )
 
         # we get the max mclk from rocm-smi --showmclkrange
-        rocm_smi_mclk = run(["rocm-smi", "--showmclkrange"], exit_on_error=True)
-        self._mspec.max_mclk = search(r"(\d+)Mhz\s*$", rocm_smi_mclk)
+        try:
+            amdsmi_init()
```
feizheng10, line 176: We should design it as a singleton, in which avoiding amdsmi_init() again.

feizheng10 (changes requested):

src/rocprof_compute_soc/soc_base.py
ywang103-amd: already done it. Now it only import functions and classes that are only used in a given file

ywang103-amd:
amd smi needs more implementations in the incoming releases to make us able to install in our existing setup procedure

PR #608, opened 2025-03-12
Debug logging during intensities calculations when no flops recorded

cfallows-amd:
Added debug log for when no flops are recorded (total_flops is 0), so AI points will not be plotted.
Removed commented out print statement that is not functional- contains nonexistent method call.


cfallows-amd:
Quick verify done against sample/vmem.hip and -V -V option.

vedithal-amd (approved):
LGTM

PR #609, opened 2025-03-13
Fix tests on MI 100

vedithal-amd:


feizheng10 (approved):

vedithal-amd:
Profiling test is failing for MI 200, I have asked @cfallows-amd to fix it

PR #610, opened 2025-03-13
Adjust roof_only list in ctest for mi200/mi300 differences

cfallows-amd:
roof_only related ctests failing because only MI300 outputs FP8 roofline PDFs. Adjusting the list such that MI200 excludes FP8 PDF expected output.

### DIFF
--- a/tests/test_profile_general.py
+++ b/tests/test_profile_general.py
@@ -116,11 +116,10 @@ ALL_CSVS_MI300 = sorted(
     ]
 )
 
-ROOF_ONLY_FILES = sorted(
+MI200_ROOF_ONLY_FILES = sorted(
     [
         "empirRoof_gpu-0_fp32_fp64.pdf",
         "empirRoof_gpu-0_int8_fp16.pdf",
-        "empirRoof_gpu-0_fp8.pdf",
         "pmc_perf.csv",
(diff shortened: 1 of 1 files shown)


cfallows-amd:
Manual ctest run passed on MI210 and MI300X systems.
Solution is temporary- tests will be reworked in upcoming tasks.

tests/test_profile_general.py
```diff
@@ -310,6 +311,14 @@ def gpu_soc():
 
 soc = gpu_soc()
 
+
+# Set ROOF_ONLY_FILES according to soc
+if soc == "MI200":
+    ROOF_ONLY_FILES = MI200_ROOF_ONLY_FILES
+else:
+    ROOF_ONLY_FILES = MI300_ROOF_ONLY_FILES
```
vedithal-amd, line 319: All the changes proposed in this PR can be made concise and readable without introducing new variables, as follows:
```
if soc == "MI200":
    # MI 200 does not support FP 8, explain why FP 8 ignore in MI 200 but not in MI 100 or MI 300
    ROOF_ONLY_FILES.remove("empirRoof_gpu-0_fp8.pdf")
```

vedithal-amd (changes requested):

PR #612, opened 2025-03-17
replace rocm-smi with amd-smi cmd

ywang103-amd:
…dsmi as temp solution for amdsmi to replace rocmsmi

### DIFF
--- a/src/utils/specs.py
+++ b/src/utils/specs.py
@@ -35,7 +35,6 @@ from math import ceil
 from pathlib import Path as path
 
 import pandas as pd
-import yaml
 
 import config
 from utils.tty import get_table_string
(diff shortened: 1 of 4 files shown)


src/utils/specs.py
```diff
@@ -35,10 +35,12 @@
 from pathlib import Path as path
 
 import pandas as pd
+import yaml
```
feizheng10: why>?

feizheng10 (changes requested):
Update changelog accordingly please

src/utils/specs.py
```diff
@@ -628,7 +637,7 @@ def run(cmd, exit_on_error=False):
         )
 
     if exit_on_error:
-        if cmd[0] == "rocm-smi":
+        if cmd[0] == "rocm-smi" or cmd[0] == "amd-smi":
             if p.returncode != 2 and p.returncode != 0:
                 console_error("No GPU detected. Unable to load rocm-smi")
```
feizheng10: no rocm-smi at all

src/utils/specs.py
ywang103-amd: was firstly thinking about using ymal to parse output of "amd-smi static". But made implementation of regular expression later on. I have moved it in the latest commit.

src/utils/specs.py
ywang103-amd: updated it in the latest commit

feizheng10 (approved):

PR #624, opened 2025-03-22
Datatype selection option for roofline

cfallows-amd:
Added command line option to specify which datatype(s) to capture into the roofline PDF(s)
All datatypes are still collected by roofline call if applicable, but only specific datatypes are plotted into PDF outputs
Default is FP32

Reworked roofline calls and plotting to be general enough such that any new datatypes added into rocm-amdgpu-bench can easily be reflected in rocprof-compute with simple modifications in roofline_calc.py

Adjusted ctest to reflect expected default pdf outputs from roofline


cfallows-amd:
Ran ctest and manual testing on MI210 and MI300X systems.

feizheng10 (approved):
LG. Changelog please.
And I guess there will some doc work to do:)

src/argparser.py
```diff
@@ -367,6 +367,19 @@ def validate_block(value):
         action="store_true",
         help="\t\t\tInclude kernel names in roofline plot.",
     )
+
+    roofline_group.add_argument(
+        "-D",
+        "--datatype",
```
feizheng10: I'd suggest to use "--roofline-data-type". In the future, we need this for post-analysis as well.

feizheng10:
User still could do combo of multiple data types in one plot, right?

src/argparser.py
cfallows-amd: Agree, will make change.

cfallows-amd:
Yes, combo plot occurs if more than one of a type (FP vs Int) gets selected, there will be more than one type on the graph. Int and FP graphs will always be separated because of AI for FP. Docs will highlight that one run of roofline will collect all data, so there is still an option to generate separate plots for each type off of the same data collection run if there exists a workload directory already populated. I will work with docs team to update the roofline section today. Thanks

PR #626, opened 2025-03-22
Fix incorrect logging in `mi_gpu_spec.py`

coleramos425:
When testing the latest `develop` branch I found that #581 uses `logging.info()` as opposed to our dedicated `console_log()` to print. If we don't use the dedicated helpers, these messages will be excluded from rocprof-computes log file and will stick out from our properly formatted outputs

image

I see @xuchen-amd did this to avoid a circular dependency in src/utils/utils.py. I propose we move the logging helpers to the src/utils/logger.py module to avoid this circular dependency issue and ensure all file can utilize our logging helpers

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,18 +2,18 @@ default_stages: [pre-commit]
 fail_fast: true
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v2.3.0
+    rev: v5.0.0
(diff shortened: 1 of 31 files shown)


feizheng10 (approved):
Thanks for catching it!

vedithal-amd (approved):
Thanks for the fix @coleramos425 , me and Xuan were also trying to fix this, but eliminating the circular dependency is the best solution

PR #627, opened 2025-03-24
fix L2 missing metrics' values with rocprofv3 that uses TCC _sum counters

ywang103-amd:


feizheng10 (approved):
This needs an entry in the  changelog

PR #629, opened 2025-03-25
fix crash for rocprofv1

ywang103-amd:


src/utils/utils.py
```diff
@@ -52,6 +52,10 @@ def is_tcc_channel_counter(counter):
     return counter.startswith("TCC") and counter.endswith("]")
 
 
+def using_v1():
+    return "ROCPROF" in os.environ.keys() and os.environ["ROCPROF"].endswith("rocprof")
+
+
 def using_v3():
     return "ROCPROF" in os.environ.keys() and "rocprofv3" in os.environ["ROCPROF"]
```
feizheng10, line 60: Please fix here as well.
return "ROCPROF" in os.environ.keys() and os.environ["ROCPROF"].endswith("rocprofv3")

feizheng10 (approved):

src/utils/utils.py
ywang103-amd: added in another PR

PR #631, opened 2025-03-25
fix the crash related to agent id in rocprofv3

ywang103-amd:
…alue to string


src/utils/utils.py
```diff
@@ -491,6 +491,14 @@ def v3_counter_csv_to_v2_csv(counter_file, agent_info_filepath, converted_csv_fi
         values="Counter_Value",
     ).reset_index()
 
+    console_debug(
+        "The type of Agent ID from counter csv file is {}".format(
+            result["Agent_Id"].dtype
+        )
+    )
+    if result["Agent_Id"].dtype == "object":
```
feizheng10, line 500: Please add comments something like:
NB: Agent_Id is int in older rocporfv3, now switched to string with prefix "Agent ". We need to make sure handle both cases.

ywang103-amd: added the comments

feizheng10 (approved):

PR #634, opened 2025-03-25
Enable non-functional performance counters for gfx950.

xuchen-amd:
- Added TCP latency counters
- Added TA wavefront counters
- Added TCC_EA0 stall counters


vedithal-amd (changes requested):
SQ_INSTS_VALU_MFMA_F8 is supported by gfx950 per https://github.com/ROCm/rocprofiler-sdk/blob/amd-mainline/source/lib/rocprofiler-sdk/counters/yaml/counter_defs.yaml , so why did we remove MFMA-F8 derived metric?

We are using TA_{FLAT/BUFFER}_COALESCABLE_WAVEFRONT_sum counter, but this is not supported, there is no such counter with this name

PR #638, opened 2025-03-26
Support host-trap PC Sampling on CLI (beta version)

feizheng10:
Support host-trap PC Sampling, CLI only, rocprofv3 only

Design and implementation details:
- Design pc_sampling block similar to any other regular panel/df with specific func to handle data and presentation separately. The block index is 21.
-  Add one arg "--pc-sampling-interval" to set  pc sampling interval, default to 1 micro second.
- As a beta version, force user to explicitly use "-b 21" to enable the feature.
- For post-analysis, if user doesn't select specific kernel, we will group, sort, and pull pc sampling info from rocprofv3 csv output  directly. For single kernel selected, we parse pc sampling info from rocprofv3 json records.

All other potential improvements and corner cases would go for incoming works.


coleramos425:
Already so much on rocprof-compute team's plate, but not being able to decipher false positives from these unit tests is concerning. Failing formatting test is an easy fix [with help from linter](https://github.com/ROCm/rocprofiler-compute/blob/04f92b72a9b810b70417199504b257f3f780e1d6/CONTRIBUTING.md?plain=1#L39-L53).

Otherwise, until CI overhaul completes, we'll just have to be very careful in code reviews. I'll add mine later today.

src/utils/utils.py
```diff
@@ -685,6 +685,37 @@ def run_prof(
     df.to_csv(workload_dir + "/" + fbase + ".csv", index=False)
 
 
+def pc_sampling_prof(interval, workload_dir, appcmd, loglevel):
```
xuchen-amd: arg `loglevel` is not being used.

src/utils/parser.py
```diff
@@ -930,11 +939,249 @@ def apply_filters(workload, dir, is_gui, debug):
...
+                kernel_info["potential_end_offset"] = item[
+                    "kernel_code_entry_byte_offset"
+                ]
+            else:
+                kernel_info["potential_end_offset"] = sys.maxsize
+            break
+
+    # print("kernel_info", kernel_info)
+
+    pc_sample_host_trap = search_key_in_json(file_name, "pc_sample_host_trap")
```
xuchen-amd, line 1073: Should we add sanity check on `pc_sample_host_trap`?
Or exception handling in `search_pc_sampling_record()`?

feizheng10: Thanks for the suggestion, fixed in https://github.com/ROCm/rocprofiler-compute/pull/638/commits/7652f306fad485b3f84e69a94ee7ba38f6167587

src/utils/utils.py
feizheng10: Fixed in https://github.com/ROCm/rocprofiler-compute/pull/638/commits/7652f306fad485b3f84e69a94ee7ba38f6167587

src/argparser.py
```diff
@@ -324,6 +324,15 @@ def validate_block(value):
         help="\t\t\tSet the format of output file of rocprof.",
     )
 
+    profile_group.add_argument(
+        "--pc-sampling-interval",
+        required=False,
+        metavar="",
+        dest="pc_sampling_interval",
+        default=1,
+        help="\t\t\tSet the interval of pc sampling in microsecond.",
```
coleramos425:
```suggestion
        help="\t\t\tSet the interval of pc sampling in microsecond (DEFAULT: 1).",
```
In other options, we state the default value in help print out

src/utils/parser.py
```diff
@@ -23,16 +23,20 @@
...
 from pathlib import Path
 
 import astunparse
 import numpy as np
 import pandas as pd
 
 from utils import schema
-from utils.logger import console_error, console_warning, demarcate
+from utils.logger import demarcate
+from utils.utils import console_debug, console_error, console_warning
```
coleramos425: Based on #626, logging helper (console_debug, console_error, etc.) now live in `utils.logger` file. Maybe you just need a rebase?

coleramos425 (approved):
A few comments. Otherwise, LGTM

src/utils/parser.py
feizheng10: Thanks, I thought I merged. But seems I messed up.
Fixed with 22f00b27

src/argparser.py
feizheng10: Fixed with 22f00b27. I mixed C++ parser and python:)

feizheng10:
CI passed, merge.

PR #639, opened 2025-03-26
fix ip block test by changing ways of extracting agent id

ywang103-amd:
…roken unit test's ip bloack case


src/utils/utils.py
```diff
@@ -451,7 +451,16 @@ def v3_counter_csv_to_v2_csv(counter_file, agent_info_filepath, converted_csv_fi
         )
     )
     if result["Agent_Id"].dtype == "object":
-        result["Agent_Id"] = result["Agent_Id"].str.extract("(\d+)").astype("int64")
+
+        def extract_id(agent_id):
+            """Function to extract the id number using regular expression"""
```
feizheng10: It should be simple like:
result["Agent_Id"] = result['Agent_Id'].str.contains(r'^Agent \d+', regex=True).astype("int64")

ywang103-amd: Modified to one line of regular expression with exception catch. Tested in branch of PR 633 and it works

feizheng10 (approved):

PR #642, opened 2025-03-26
Add datatypes for roofline profiling

cfallows-amd:
Rebuild of rocm-amdgpu-bench roofline binaries for MI200/MI300 systems with rocm6.
Added datatype options to roofline feature.


cfallows-amd:
Manually tested various datatype combinations with new roofline bins on MI210 and MI300X systems

src/utils/roofline_calc.py
```diff
@@ -44,13 +44,13 @@
 # SUPPORTED_DATATYPES table is based on datatype support in rocm-amdgpu-bench repository
 # Indicates which datatypes per gpu arch can be generated by the roofline binary
```
feizheng10, line 45: Might want to add NB: this should be moved to a centrelized place later.
I believe it will not only for roofline:)

feizheng10 (approved):

PR #645, opened 2025-03-27
Read Accum_VGPR_Count from rocprof output

benrichard-amd:
Recent rocprof versions provide the accumulation VGPR count.

Tested against rocprofv3 with and without accum_vgpr_count
Tested on workloads with/without accumulation VGPR usage.


feizheng10 (approved):

feizheng10:
Maybe mention it in changelog?

coleramos425 (approved):

PR #646, opened 2025-03-31
Fix bandwidth plotting on roofline outputs

cfallows-amd:
Fix roofline plots for datatypes that have peakVALU only. Check for highest roofline to plot the bandwidth lines to proper height, don't rely on existence of peakMFMA for every datatype.


cfallows-amd:
Verified all individual datatypes and some combination datatype plots on MI210 system.
[BADempirRoof_gpu-0_I32.pdf](https://github.com/user-attachments/files/19541337/BADempirRoof_gpu-0_I32.pdf)
[GOODempirRoof_gpu-0_I32.pdf](https://github.com/user-attachments/files/19541333/GOODempirRoof_gpu-0_I32.pdf)

PR #649, opened 2025-04-01
Weekly rebase liangdin-test on top of amd-mainline

vedithal-amd:


vedithal-amd:
This commit which was merged to develop was deleted from develop.
I was testing some github workflow changes.

PR #651, opened 2025-04-01
Add dependent script path to system path in rocprof-compute

raramakr:
In wheel environment, rocprof-compute in bin folder is not a soft link. For executing rocprof-compute from bin folder, the system path should also have the dependency script paths. Added the same

### DIFF
--- a/src/rocprof-compute
+++ b/src/rocprof-compute
@@ -28,6 +28,7 @@
 
 import re
 import sys
+import os
 
 # import logging
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


src/rocprof-compute
```diff
@@ -38,9 +39,20 @@ try:
...
     from utils.logger import console_error
 except ImportError as e:
-    # print("Failed to import required modules: " + str(e))
-    pass
-
+    # In wheel package, softlink is not supported.
+    # rocprof-compute will get installed in bin and libexec/rocprofiler-compute.
+    # When invoked from bin folder, the extra paths are required to import the dependent scripts
+    try:
+        current_path = os.path.dirname(os.path.abspath(__file__))
```
feizheng10, line 46: Just want to double check: you expect to use current path, or maybe the root path of ROCm?

feizheng10 (approved):

src/rocprof-compute
raramakr: Expecting current path and use the same to find the relative path 
{current_path}/../libexec/rocprofiler-compute

PR #656, opened 2025-04-02
chip logic improvements

xuchen-amd:
- add perfmon config as static yaml file
- avoid perfmon config in-file hardcode
- enhance memory partition information
- refactor `total_xcds()`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -247,8 +247,9 @@ add_test(
 add_test(
     NAME test_profile_section
     COMMAND
-        ${Python3_EXECUTABLE} -m pytest -m section --junitxml=tests/test_profile_misc.xml
-        ${COV_OPTION} ${PROJECT_SOURCE_DIR}/tests/test_profile_general.py
+        ${Python3_EXECUTABLE} -m pytest -m section
+        --junitxml=tests/test_profile_section.xml ${COV_OPTION}
+        ${PROJECT_SOURCE_DIR}/tests/test_profile_general.py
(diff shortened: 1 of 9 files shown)


src/utils/mi_gpu_spec.yaml
```diff
@@ -77,8 +77,11 @@ mi_gpu_spec:
                   qpx: null
                   cpx: null
               memory_partition_mode:
-                nps4: [tpx]
-                nps1: [spx, tpx]
+                nps4:
```
vedithal-amd: Could you clarify what do these integers mean and where is it being used

src/utils/perfmon_config.yaml
```diff
@@ -0,0 +1,84 @@
+# --------------------------------------------------------------------------------
```
vedithal-amd: Why cant this go in mi_gpu_spec.yaml file

src/utils/mi_gpu_spec.yaml
xuchen-amd: > Could you clarify what do these integers mean and where is it being used

The integers are corresponding # of HBM memory regions for the compute partition mode. This tells the user the supported partition combinations for the system they are using.

vedithal-amd: Should we use it in the hbm channels calculation?

src/utils/perfmon_config.yaml
vedithal-amd: lets consolidate all hw specs in one file

PR #663, opened 2025-04-07
Add mi325 specs.

xuchen-amd:
- [x] ran ctest locally on mi325x


ywang103-amd (approved):

feizheng10 (commented):
I think we need changelog update

PR #668, opened 2025-04-11
configure rocprofv3 as default for unit test

ywang103-amd:


xuchen-amd (commented):
Should we support all 3 rocprof versions (v1, v2, v3) for socs supported?

Also changelog update? :)

ywang103-amd:
@xuchen-amd I have updated changelog. @feifei14119 For the current version, I set env variable to rocprofv3 but still keep old file lists that supports v1 and v2 in the actual implementation. I wonder if we want to preserve them.

feizheng10 (approved):

PR #669, opened 2025-04-11
Fix rpath checks during RPM generation on RHEL10

cfallows-amd:
Invalid rpath on roofline binaries reported during build testing for new RHEL10 addition, removed rpaths to prevent rpath check failures.


cfallows-amd:
Tested ubuntu mi300 binary locally generating roofline. Tested by Jenifer through pipeline- reported rocprof-compute passing using the above binaries.

dgaliffiAMD (approved):

PR #674, opened 2025-04-21
Add test for gfx942 number of xcds.

xuchen-amd:
- Improve mi gpu spec structure: use singleton class to capture gpu specs as an instance and allow easier unit test integration
- Add num_xcd unit tests: spec_class and cli_output verification
- [x] ctest passed locally on mi300x_a1

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,5 +71,5 @@ markers = [
 	"kernel_verbose",
 	"serial",
 	"L1_cache",
-  "num_xcds",
+	"num_xcds",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,4 +71,5 @@ markers = [
 	"kernel_verbose",
 	"serial",
 	"L1_cache",
+  "num_xcds",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,5 +71,6 @@ markers = [
 	"kernel_verbose",
 	"serial",
 	"L1_cache",
-	"num_xcds",
+	"num_xcds_spec_class",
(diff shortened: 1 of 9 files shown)


pyproject.toml
```diff
@@ -71,4 +71,5 @@ markers = [
 	"kernel_verbose",
 	"serial",
 	"L1_cache",
+  "num_xcds",
```
cfallows-amd: indent formatting (checker is down right now)

tests/test_gpu_specs.py
```diff
@@ -0,0 +1,136 @@
...
+    return gpu_arch
+
+
+@pytest.mark.num_xcds
+def test_num_xcds(binary_handler_profile_rocprof_compute):
+    arch = gpu_arch()
+
+    # 1. Check if gfx942 soc
+    if not arch or "gfx942" not in arch.lower():
+        pytest.skip("Skipping num xcds test for non-fx942 socs.")
```
cfallows-amd: minor typo - gfx942

skyreflectedinmirrors:
Is it not possible to simply import `generate_machine_specs` from `specs.py` and simply test the field in `MachineSpecs`? 

There's a million ways a string parsing test like this could fail.

xuchen-amd:
> Is it not possible to simply import `generate_machine_specs` from `specs.py` and simply test the field in `MachineSpecs`?
> 
> There's a million ways a string parsing test like this could fail.

Good point, the rationale for testing `rocprof-compute -s` is to  verify information visible to the user.
Will add test to verify the spec parser/generate_machine_spec.

feizheng10 (approved):

PR #681, opened 2025-04-28
Add chip specs

xuchen-amd:
- Introduce perfmon configs as static data using specs yaml file.
- Add `mi355` specs to specs yaml file.
- Update specs yaml file with latest compute partition and memory partition specs
- Add `gfx942 high frequency (hf)` variants specs to specs yaml file.
- Add `gfx942 hf` variants to gpu tests.
- Add backup detection methods (using `marketing name`, `vbios name`, `product name`) for `gpu model` in addition to `chip ID`.
- Add backup detection methods (using `ACCELERATOR PARTITION` in `amd-smi static`, `rocm-smi --showcomputepartition`) for `compute partition` in addition to searching `COMPUTE PARTITION` in `amd-smi static`.
- Use default compute partition setting of `SPX:8` when no `compute partition` or `gpu model` is determined.
- Improve error messages and failure checks.

- [x] local ctest run on mi308x


src/utils/specs.py
```diff
@@ -674,9 +674,13 @@ def total_sqc(archname, numCUs, numSEs):
 
 
 def total_l2_banks(archname, L2Banks, compute_partition):
-    xcds = mi_gpu_specs.get_num_xcds(archname, compute_partition)
-    totalL2Banks = L2Banks * xcds
-    return totalL2Banks
+    total_l2_banks = None
+    xcd_count = mi_gpu_specs.get_num_xcds(archname, compute_partition)
+
+    if xcd_count is not None:
```
vedithal-amd: Should we add a console warning here

src/utils/specs.py
```diff
@@ -674,9 +674,13 @@ def total_sqc(archname, numCUs, numSEs):
...
+        total_l2_banks = L2Banks * xcd_count
```
vedithal-amd: Make sure num xcd and total l2 banks are being calculated correctly

src/rocprof_compute_soc/soc_base.py
```diff
@@ -186,12 +186,72 @@ def populate_mspec(self):
...
+        ]
+
+        gpu_model = None
+        for method in detection_methods:
+            console_warning(f"Determining GPU model using {method['name']}.")
+            gpu_model = search(method["pattern"], amd_smi_static)
+            if gpu_model:
+                break
+
+        if not gpu_model:
```
vedithal-amd, line 231: If we dont know the gpu model but know the gfx architecture, can we still proceed with the first model in that gfx architecture? Can we make this assumption while emitting a warning for the user?

vedithal-amd (dismissed):
Minor comments

src/rocprof_compute_soc/soc_base.py
xuchen-amd: This ask to use gfx arch got me thinking about finding all the 1-to-1 mapping of gfx arch to compute partition pairs available in the specs.yaml, and whenever available, we use gfx arch (instead of gpu model) to determine the compute partition: [def _populate_gpu_arch_to_compute_partition_dict()](https://github.com/ROCm/rocprofiler-compute/blob/10c3c1ecf6394c552e690f30bdbe5bdec62e40b4/src/utils/mi_gpu_spec.py#L137)
In the case this 1:1 relation is not available, it would also be difficult to make assumptions of the model using gfx arch alone. e.g., gfx942 holds two `compute_partition_mode`.

cc. @feizheng10

vedithal-amd: If we can't get the exact gpu model but only the gfx architecture, we could assume the first gpu model in that gfx architecture as the default. Obviously, this is an approximation, but i am not sure if there is anything else we can do here.

src/rocprof_compute_soc/soc_base.py
```diff
@@ -186,12 +186,72 @@ def populate_mspec(self):
...
+        Detects the GPU model using various identifiers from 'amd-smi static'.
+        Falls back through multiple methods if the primary method fails.
+        """
+
+        from utils.specs import run, search
+
+        amd_smi_static = run(["amd-smi", "static", "--gpu=0"], exit_on_error=True)
+
+        # Purposely search for patterns without variants suffix to try and match a known GPU model.
+        detection_methods = [
```
feizheng10, line 212: Maybe add a comment here: we should check those through amd-smi python api in the future.

feizheng10 (approved):

xuchen-amd:
CI failures not relating to this PR:
- gfx942 staging - test_profile_ipblocks failed
- gfx942 mainline - Unable to resolve path to rocprofv3 binary
- gfx90a staging - test_profile_ipblocks failed
- gfx90a mainline - Unable to resolve path to rocprofv3 binary

PR #683, opened 2025-04-29
Code fix for installation by package manager

prbasyal-amd:
Fix for the following ticket:
https://ontrack-internal.amd.com/browse/SWDEV-529961

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -11,3 +11,6 @@ build:
 python:
   install:
   - requirements: docs/sphinx/requirements.txt
+
+sphinx:
+   configuration: docs/conf.py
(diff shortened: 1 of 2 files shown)


yugang-amd (approved):
The PR looks good to me, so I approved it.  

However I did notice a build failure caused by missing parameters in `.readthedocs.yaml`. Based on a comparison with the `ROCm-internal` YAML configuration (https://github.com/ROCm/ROCm-internal/blob/develop/.readthedocs.yaml), I believe adding the following lines to `.readthedocs.yaml` should resolve the issue:

```
sphinx:
   configuration: docs/conf.py
```
  
I can raise a separate PR to address this if needed. Thanks!

For more information about the fix, see https://about.readthedocs.com/blog/2024/12/deprecate-config-files-without-sphinx-or-mkdocs-config/

Issue #689, opened 2025-04-30
KeyError: `Grid_Size' when no kernel are launched


etiennemlb:
```
rocprof-compute profile -n dummy_mi250x --roof-only --device 0 --mem-level HBM -- sleep 1
```

I get 
```
Empty DataFrame
Columns: [Dispatch_ID, Kernel_Name]
Index: []
Traceback (most recent call last):
  File "/rocprofiler-compute/bin/rocprof-compute", line 156, in <module>
    main()
  File "/rocprofiler-compute/bin/rocprof-compute", line 144, in main
    rocprof_compute.run_profiler()
  File "/rocprofiler-compute/libexec/rocprofiler-compute/utils/utils.py", line 53, in wrap_function
    result = function(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/rocprofiler-compute/libexec/rocprofiler-compute/rocprof_compute_base.py", line 286, in run_profiler
    profiler.post_processing()
  File "/rocprofiler-compute/libexec/rocprofiler-compute/utils/utils.py", line 53, in wrap_function
    result = function(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/rocprofiler-compute/libexec/rocprofiler-compute/rocprof_compute_profile/profiler_rocprof_v1.py", line 100, in post_processing
    self.join_prof()
  File "/rocprofiler-compute/libexec/rocprofiler-compute/utils/utils.py", line 53, in wrap_function
    result = function(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/rocprofiler-compute/libexec/rocprofiler-compute/rocprof_compute_profile/profiler_base.py", line 116, in join_prof
    key = _df.groupby(["Kernel_Name", "Grid_Size"]).cumcount()
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/rocprofiler-compute/python-libs/pandas/core/frame.py", line 9183, in groupby
    return DataFrameGroupBy(
           ^^^^^^^^^^^^^^^^^
  File "/rocprofiler-compute/python-libs/pandas/core/groupby/groupby.py", line 1329, in __init__
    grouper, exclusions, obj = get_grouper(
                               ^^^^^^^^^^^^
[...]

PR #694, opened 2025-05-02
Avoid crash when profiling data not generated

benrichard-amd:
Addresses #689 

rocprof-compute would crash if the program did not have any kernel launches

- Handle case where program has no kernel launches/rocprof has missing data

- Improve error messages

- Skip roofline if no profiling data was collected (e.g. pmc_perf.csv does not exist).


cfallows-amd (approved):

benrichard-amd:
Don't merge. I think there are other `soc_gfx*.py` files that need  updating.

PR #695, opened 2025-05-03
Implement interface to rocprofiler sdk

vedithal-amd:
Implement interface to rocprofiler sdk

  * Add profiler mode = rocprofiler-sdk which will not invoke rocprofv3 but use rocprofiler-sdk directly, this can be selected by setting env. var. ROCPROF=rocprofiler-sdk
  
  * Add runtime option --rocprofiler-sdk-library-path to use custom version of rocprofiler sdk library
      * Add --rocprofiler-sdk-library-path conftest option for tests
  
  * Setup appropriate environment variables to inject rocprofiler sdk code to user command
      * Add env. vars. for counter collection and filtering
      * Add env. vars. for pc sampling
   
  * Use python bindings to list counters supported by rocprofiler sdk

  * Update changelog

### DIFF
--- a/src/rocprof_compute_profile/profiler_rocprofiler_sdk.py
+++ b/src/rocprof_compute_profile/profiler_rocprofiler_sdk.py
@@ -1,0 +1,126 @@
+##############################################################################bl
+# MIT License
+#
+# Copyright (c) 2021 - 2025 Advanced Micro Devices, Inc. All Rights Reserved.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+##############################################################################el
+
+import os
+import shlex
+from pathlib import Path
+
+from rocprof_compute_profile.profiler_base import RocProfCompute_Base
+from utils.logger import console_error, console_log, demarcate
+
+
+class rocprofiler_sdk_profiler(RocProfCompute_Base):
+    def __init__(self, profiling_args, profiler_mode, soc, supported_archs):
+        super().__init__(profiling_args, profiler_mode, soc, supported_archs)
+        self.ready_to_profile = (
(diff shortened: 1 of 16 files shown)


vedithal-amd:
Note that only mainline azure pipeline changes are failing (staging passes) with the error "rocprofv3 --list-avail failed due to rocprofv3 not found" cc @danielsu-amd.

src/rocprof_compute_profile/profiler_rocprof_sdk.py
```diff
@@ -0,0 +1,126 @@
...
+        super().__init__(profiling_args, profiler_mode, soc, supported_archs)
+        self.ready_to_profile = (
+            self.get_args().roof_only
+            and not Path(self.get_args().path).joinpath("pmc_perf.csv").is_file()
+            or not self.get_args().roof_only
+        )
+
+    def get_profiler_options(self, fname, soc):
+        app_cmd = shlex.split(self.get_args().remaining)
+        rocm_libdir = str(Path(os.environ["ROCPROF_SDK_LIBRARY_PATH"]).parent)
```
feizheng10: Hit a topic I was trying not to talk now:)
First, the name of original env var ROCPROF is not a good one. Each component should have its own env var set starts with its project prefix, in this case,  technically, something like ROCPROF_COMPUTE_***. The *** could be 
 PROFILER, COLLECTOR, ENGINE, or something else. We should never use ROCPROF because we never know it will be used for rocprof itself or some other project. Using ROCPROF_SDK_LIBRARY_PATH would extend the problem more. However, with backward compatibility, I hesitate to suggest to change it now.
   
I see your point to use "rocprofsdk", and with separate PATH config. The name here is short and clear. The PATH is only for the sdk path. Still, officially we should only use the name "rocprofiler-sdk" or "Rocprofiler-sdk". Any unofficial variant will lead more confusion and problem later.

Generally speaking, for this kind config, there are 3 ways to config: env var, config file, runtime option. We need to be careful to evaluate the pros and cons. For instance, technically, with env var, we need a set like ROCPROF_COMPUTE_PROFILER, ROCPROF_COMPUTE_AMD_SMI, ROCPROF_COMPUTE_ROCM_INFO. 
So far,  my preference might be runtime option with providing rocm_root and profiler option.

src/utils/utils.py
```diff
@@ -575,6 +595,31 @@ def v3_counter_csv_to_v2_csv(counter_file, agent_info_filepath, converted_csv_fi
     result.to_csv(converted_csv_file, index=False)
 
 
+def parse_text(text_file):
+    def process_line(line):
```
feizheng10, line 601: always add fun comments please

src/rocprof_compute_soc/analysis_configs/gfx906/2100_pc_sampling.yaml
```diff
@@ -5,5 +5,6 @@ Panel Config:
   data source:
     - pc_sampling_table:
         id: 2101
+        title: PC Sampling
```
feizheng10: Thanks for catching up this. Next time, please put any small fixes in separate PR, for any potential code reverting.

feizheng10:
And changelog please

src/rocprof_compute_profile/profiler_rocprof_sdk.py
vedithal-amd: I will add runtime option to specify rocprofiler sdk library path

src/rocprof_compute_soc/analysis_configs/gfx906/2100_pc_sampling.yaml
vedithal-amd: I will move changes not related to rocprofiler sdk interface to another PR

vedithal-amd:
> And changelog please

@feizheng10 , this is a dependency management related change and will not affect the user experience or functionality of the tool, do we need a changelog statement for this change?

src/rocprof_compute_profile/profiler_rocprofiler_sdk.py
```diff
@@ -0,0 +1,126 @@
+##############################################################################bl
+# MIT License
+#
+# Copyright (c) 2021 - 2025 Advanced Micro Devices, Inc. All Rights Reserved.
```
feizheng10, line 4: 2025 only

feizheng10 (approved):
Thanks!

PR #703, opened 2025-05-14
Update illegal character check on profile name

benrichard-amd:
- Remove restriction on dot (.) and dash (-) in profile names
- Add restriction on forward slash (/) in profile name
- Move profile name check to beginning of program before directory is created. Previously we were creating the directory and then checking the name.

### DIFF
--- a/src/rocprof_compute_base.py
+++ b/src/rocprof_compute_base.py
@@ -252,6 +252,9 @@ class RocProfCompute:
         elif self.__args.name is None:
             sys.exit("Either --list-name or --name is required")
 
+        if self.__args.name.find("/") != -1:
+            console_error("'/' not permitted in profile name")
+
         # Deprecation warning for hardware blocks
         if [
             name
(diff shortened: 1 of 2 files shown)


coleramos425:
Before we merge, @feizheng10, have we officially deprecated Grafana? For any HPC users who still use Grafana, this change has potential to cause them problems. The condition was originally implemented to comply with MongoDB naming constraints.

My opinion is officially deprecate the feature (i.e. removing all instances in repo) before introducing a breaking change

vedithal-amd (approved):
LGTM

A quick way to deprecate database support would be:

1. Add deprecated notice in help message at [https://github.com/ROCm/rocprofiler-compute/blob/a8588780804d69cf86c2d2c59c924cc71aa927eb/src/argparser.py#L409](https://github.com/ROCm/rocprofiler-compute/blob/a8588780804d69cf86c2d2c59c924cc71aa927eb/src/argparser.py#L409)

2. If database mode is triggered, exit with deprecated error here[ https://github.com/ROCm/rocprofiler-compute/blob/a8588780804d69cf86c2d2c59c924cc71aa927eb/src/rocprof-compute#L159  ]( https://github.com/ROCm/rocprofiler-compute/blob/a8588780804d69cf86c2d2c59c924cc71aa927eb/src/rocprof-compute#L159  )


A full fledged effort to remove everything database related can be done as part of a separate PR

PR #705, opened 2025-05-14
Check mode during soc init

cfallows-amd:
Prevent roofline object creation if there is no mode (profile, analyze, database) provided- in short, skip if only printing specs.


cfallows-amd:
Not caught during PR other day because nodes were down- caught by Daniel while enabling nodes and testing manually.
Manually tested changes by running ctest on top of latest develop branch on MI300X system. Changes also apply to gfx940 and gfx941 but no systems available to test those. MI350 series has issues with ctest at the moment, but applied the change there as well.

cfallows-amd:
rocprofiler_compute_test_gfx942/gfx90a_mainline failing for rocprofv3 purposes
rocprofiler_compute_test_gfx942/gfx90a_staging failing for ip_blocks TCC test

Above tests unrelated to change made. Specific tests originally failing due to this issue are now passing in the above tests.

xuchen-amd (approved):

PR #707, opened 2025-05-16
Fixed duplicate keys in analysis_configs yamls

jamessiddeley-amd:
Duplicate keys found in yaml files, found using formatting tool.

### DIFF
--- a/src/rocprof_compute_soc/analysis_configs/gfx940/1000_compute-unit-instruction-mix.yaml
+++ b/src/rocprof_compute_soc/analysis_configs/gfx940/1000_compute-unit-instruction-mix.yaml
@@ -281,4 +281,9 @@ Panel Config:
             min: MIN((SQ_INSTS_VALU_MFMA_F64 / $denom))
             max: MAX((SQ_INSTS_VALU_MFMA_F64 / $denom))
             unit: (instr + $normUnit)
-          # TODO: Fix baseline comparision logic to handle non existent metrics, then
+          MFMA-F6F4:
+            avg: None
+            min: None
+            max: None
+            unit: (instr + $normUnit)
+            tips:
--- a/src/rocprof_compute_soc/analysis_configs/gfx940/1000_compute-unit-instruction-mix.yaml
+++ b/src/rocprof_compute_soc/analysis_configs/gfx940/1000_compute-unit-instruction-mix.yaml
@@ -281,11 +281,4 @@ Panel Config:
             min: MIN((SQ_INSTS_VALU_MFMA_F64 / $denom))
             max: MAX((SQ_INSTS_VALU_MFMA_F64 / $denom))
             unit: (instr + $normUnit)
-          # TODO: Fix baseline comparision logic to handle non existent metrics, then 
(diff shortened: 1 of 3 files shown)


src/rocprof_compute_soc/analysis_configs/gfx940/1000_compute-unit-instruction-mix.yaml
```diff
@@ -281,11 +281,4 @@ Panel Config:
...
             unit: (instr + $normUnit)
-          # TODO: Fix baseline comparision logic to handle non existent metrics, then 
-          MFMA-F6F4:
-            avg: None
-            min: None
-            max: None
-            unit: (instr + $normUnit)
-            tips:
-            tips:
+          # TODO: Fix baseline comparision logic to handle non existent metrics, then
```
xuchen-amd: the comment can be removed too :)

vedithal-amd: We should not have None values in our analysis configuration files. The only reason we have these None values is because the analysis baseline comparison logic is not smart enough to automatically set non existent metrics as None. This comment is a reminder to fix that logic. If you are removing None values then the comment can be removed too. Thank you for removing duplicate None values :)

src/rocprof_compute_soc/analysis_configs/gfx940/1000_compute-unit-instruction-mix.yaml
```diff
@@ -281,11 +281,3 @@ Panel Config:
             min: MIN((SQ_INSTS_VALU_MFMA_F64 / $denom))
             max: MAX((SQ_INSTS_VALU_MFMA_F64 / $denom))
             unit: (instr + $normUnit)
-          # TODO: Fix baseline comparision logic to handle non existent metrics, then 
-          MFMA-F6F4:
```
vedithal-amd, line 285: I dont think this is duplicate so why are we removing this? Maybe my understanding of duplicate is wrong.

vedithal-amd (changes requested):

src/rocprof_compute_soc/analysis_configs/gfx940/1000_compute-unit-instruction-mix.yaml
xuchen-amd: > We should not have None values in our analysis configuration files. The only reason we have these None values is because the analysis baseline comparison logic is not smart enough to automatically set non existent metrics as None. This comment is a reminder to fix that logic. If you are removing None values then the comment can be removed too. Thank you for removing duplicate None values :)

Good point. We should also include this comment in the source code (such as in [def show_all()](https://github.com/ROCm/rocprofiler-compute/blob/41dd4aab907d043bd9895229c17343caa8f8d7a3/src/utils/tty.py#L64), or somewhere more logic-appropriate) in addition to the yaml files to better track TODOs.

vedithal-amd (approved):
LGTM

xuchen-amd (approved):

PR #709, opened 2025-05-20
Add F4 F6 to roofline for MI350 series

cfallows-amd:
Add F4 and F6 roofline benchmarking - using mfma scale f8f6f4 builtin.
Add F6F4 perfmon event- event is combined counter for F4 F6.
Only for mi350 series.


cfallows-amd:
newplot (2)

src/argparser.py
xuchen-amd, line 1: Very minor, but would it help to have `choices` as a static information (only mentioning because this information is used multiple times)?

xuchen-amd (approved):

feizheng10 (approved):

src/argparser.py
cfallows-amd: Will make this a follow-up task to prevent blocking this submission- will look at adjusting roofline-data-type option choices in conjunction with a general datatype availability list to reference.

cfallows-amd:
Pipeline testing failing due to smi no gpu loading and rocprof unavailable. Not related to changes.
This PR has been manually tested roofline and also passes ctest- on MI355X system with rocm6.5.

PR #710, opened 2025-05-22
Refactor tty for better modularity.

xuchen-amd:
A refactoring attempt on tty `def show_all()`, the goal is to
- encapsulate Panel related functions using OOP
- add exception/error handling for safer data access
- modular designs to separate different logics


feizheng10 (approved):
LGTM.
I'd say we might want to restructure a bit for all our supported front-ends with one directory for each.

PR #717, opened 2025-05-26
Rename roofline bins

cfallows-amd:
Removing rocm version from binary name, not necessary to keep at this time. Name now only distinguishes between distributions binary is built on.


vedithal-amd:
How are these roofline binaries accessed from rocprofiler-compute? Do we need to make any code changes such that rocprofiler-compute accesses roofline binaries with correct name? Is there anything else which calls these roofline binaries?

feizheng10 (approved):

cfallows-amd:
> How are these roofline binaries accessed from rocprofiler-compute? Do we need to make any code changes such that rocprofiler-compute accesses roofline binaries with correct name? Is there anything else which calls these roofline binaries?

Yes needed the extra changes to detect bin, did not get pushed after commit. Pushed commit after doing manual tests on MI300X system- passed manual testing with and without ROOFLINE_BIN env var, and a ctest run.

cfallows-amd:
gfx942 staging and mainline - failing due to gpu not detected
gfx90a staging - failing in ip_blocks test for number of files outputted
gfx90a mainline - failing because cannot find rocprofv3
^ All above tests failing unrelated to this PR, passing these notes on to team. Running manually and running ctest on MI300 system for this reason to confirm bins and bin detection are still working as expected.

vedithal-amd (approved):
LGTM!

PR #723, opened 2025-05-30
Tcc new format input yaml

ywang103-amd:
fix of TCC channel counters that supports new format for -E yaml file of rocprofv3


vedithal-amd:
To add more context, this change is due to the following change in rocprofiler-sdk which is our upstream dependency: https://github.com/ROCm/rocprofiler-sdk/commit/33e43e66d3b3e73b15d81d6aa0b128a0cecb8c72

vedithal-amd (approved):
LGTM, make sure all tests pass on MI 300

PR #725, opened 2025-06-03
Roofline updates- add YAML file and add CLI plot function

cfallows-amd:
- Roofline_calc.yaml now serves as main hub for roofline equations and variables (such as supported datatypes per arch)
- cli_generate_plot() plots roofline in terminal
- Roofline bins updated- fix issue with populating headers in roofline for F4 and F6, no functional changes


src/roofline.py
```diff
@@ -242,13 +265,81 @@ def generate_plot(self, dtype, fig=None) -> go.Figure():
...
         plot_mode = "lines+text" if self.__run_parameters["is_standalone"] else "lines"
         self.__ceiling_data = constuct_roof(
             roofline_parameters=self.__run_parameters,
             dtype=dtype,
+            roof_types=self.__roofline_calc["GROUPED_DATATYPES"],
         )
+        ops_flops = "OP" if (dtype[:1] == "I") else "FLOP"  # For printing purposes
         console_debug("roofline", "Ceiling data:\n%s" % self.__ceiling_data)
 
+        #######################
```
cfallows-amd, line 280: Moving plotting AI up in the method to put it at the top of the legend.
Also removing per-datatype AI plot points, because we are just replotting them in the end (total_flops/cachedata) at this time.

src/roofline.py
```diff
@@ -333,86 +422,182 @@ def generate_plot(self, dtype, fig=None) -> go.Figure():
...
-                        mode="markers",
-                        marker_symbol=(
-                            SYMBOLS
-                            if self.__run_parameters["include_kernel_names"]
-                            else None
-                        ),
-                    )
+        return fig
+
+    def cli_generate_plot(self):
```
cfallows-amd, line 431: cli_generate_plot very similar to normal generate_plot, but does have some differences that d not allow straight copy of the work at this time.

src/utils/roofline_calc.py
```diff
@@ -35,71 +35,12 @@
 IMGNAME = "empirRoof"
 
 XMIN = 0.01
-XMAX = 1000
+XMAX = 1000000
```
cfallows-amd, line 38: XMAX as 1000 way too small with values I was getting using occupancy sample. 1M is safe for now but a todo is to gauge the length based on the farthest AI point.

src/utils/rooflines/roofline-rhel8
cfallows-amd, line 1: Binaries had an issue with the headers populated to roofline csv- was missing f4 and f6 so they were misaligned. Copied bin and tested, must have not committed the final bin copy. Has been fully corrected by grabbing the latest built bin, no functional changes to user or values.

src/utils/roofline_calc.yaml
```diff
@@ -0,0 +1,341 @@
...
+    + (df["SQ_INSTS_VALU_MFMA_MOPS_F16"][idx] * 512)
+    + (df["SQ_INSTS_VALU_MFMA_MOPS_BF16"][idx] * 512)
+    + (df["SQ_INSTS_VALU_MFMA_MOPS_F32"][idx] * 512)
+    + (df["SQ_INSTS_VALU_MFMA_MOPS_F64"][idx] * 512)
+  )
+
+total_flops_fp8: &total_flops_fp8_anchor df["SQ_INSTS_VALU_MFMA_MOPS_F8"][idx] * 512
+total_flops_f6f4: &total_flops_f6f4_anchor df["SQ_INSTS_VALU_MFMA_MOPS_F6F4"][idx] * 512
+
+total_flops:
```
cfallows-amd, line 139: Using anchors to reuse equations as needed. Yaml doesn't have ability to compound more than one anchor into a value for a single key, so total_flops needs to use a list to workaround it.
Each measurement has a specific per-arch key so that even if called in roofline methods it is still protected against using on an arch that does not support the datatype.

src/utils/roofline_calc.yaml
```diff
@@ -0,0 +1,341 @@
...
+# |-- {calculation}_additionaltypes: ...
+# |-- {calculation}:
+#         |-- {gfx_arch1}:
+#         |-- {gfx_arch2}:
+#                ...
+#         |-- {gfx_archN}:
+# ----------------------------------------
+
+
+# Variables
```
cfallows-amd, line 19: Moved datatype lists into yaml for single location to update. Contains supported datatypes per arch, and which datatypes produce VALU and MFMA values.

src/utils/roofline_calc.py
```diff
@@ -630,6 +474,6 @@ def constuct_roof(roofline_parameters, dtype):
     # ------------------
     #  Generate Roofline
     # ------------------
-    results = calc_ceilings(roofline_parameters, dtype, benchmark_data)
+    results = calc_ceilings(roofline_parameters, dtype, roof_types, benchmark_data)
```
cfallows-amd, line 477: Passing supported datatypes per arch info through to calculate the roofline ceilings

src/utils/roofline_calc.py
```diff
@@ -285,73 +224,18 @@ def calc_ai(mspec, sort_type, ret_df):
...
-                + (df["SQ_INSTS_VALU_MFMA_MOPS_F16"][idx] * 512)
-                + (df["SQ_INSTS_VALU_MFMA_MOPS_BF16"][idx] * 512)
-                + (df["SQ_INSTS_VALU_MFMA_MOPS_F32"][idx] * 512)
-                + (df["SQ_INSTS_VALU_MFMA_MOPS_F64"][idx] * 512)
-            )
-            if "FP8" in supported_dt:
-                total_flops += df["SQ_INSTS_VALU_MFMA_MOPS_F8"][idx] * 512
-            if ("FP4" in supported_dt) or ("FP6" in supported_dt):
-                total_flops += df["SQ_INSTS_VALU_MFMA_MOPS_F6F4"][idx] * 512
+            for ops in roof_eq["total_flops"][mspec.gpu_arch]:
```
cfallows-amd, line 227: Using yaml content here. Removing all equations. All arch are accounted for in terms of whether they support the data type.

cfallows-amd:
[empirRoof_gpu-0_FP32.pdf](https://github.com/user-attachments/files/20561892/empirRoof_gpu-0_FP32.pdf)
Screenshot 2025-06-02 233016

src/utils/roofline_calc.yaml
```diff
@@ -0,0 +1,341 @@
+---
```
feizheng10, line 1: Let's discuss it later:
My expectation would be: 0400_roofline.yaml for each arch. The 0400_roofline.yaml should be designed and generalized no diff than any other yaml besides specifying cli_style/tui_style. 
Each arch only enables its supported data type. No worry about duplication across archs. Eventually, Vignesh's PR will template it with Jinja.

cfallows-amd:
Per last comment, yaml file needs to be separated. Closing.

PR #726, opened 2025-06-04
Unified configuration for metrics

vedithal-amd:
* Show description of metrics during analysis
    * Use --include-cols Description show the Description column in analyze mode (this is hidden by default)
    * Remove tips field from analysis config

For example:
{62F64403-D78A-4A97-8ADD-BCF7941BB1BA}

* Align metric names in analysis config and documentation
* Add unified config utils/unified_config.yaml
* Add python script utils/split_config.py to auto generate analysis configuration and documentation metrics description
   * Add test case to ensure unified config is older than auto-generated config
   * Auto generate analysis config and documentation metrics description

* Update CONTRIBUTING.md to add instructions to build documentation assets
    * Add docker image and compose file to build documentation
* Update CHANGELOG and Documentation

* Use jinja template instead of hardcoding metric tables in documentation


tests/test_gpu_specs.py
```diff
@@ -147,7 +147,7 @@ def test_num_xcds_cli_output():
 
     # 2. Run rocprof-compute -s and grab rocprof-compute num_xcd
     proc = subprocess.run(
-        ["rocprof-compute", "-s"],
+        ["src/rocprof-compute", "-s"],
```
xuchen-amd: is `src/rocprof-compute` an expected executable path change?

src/argparser.py
```diff
@@ -670,7 +670,19 @@ def validate_block(value):
...
+    analyze_advanced_group.add_argument(
+        "--exclude-cols",
+        dest="exclude_cols",
+        metavar="",
+        nargs="*",
+        default=["Metric Description"],
+        help=(
+            "\t\tSpecify column names to exclude in cli output.\n"
+            "\t\tDefaults to 'Metric Description'.\n"
+            "\t\tTo show 'Metric Description' column, use --exclude-cols with no arguments."
```
xuchen-amd: I'm a little bit confused by the help menu.
What is the difference between `--exclude-cols` default and no args when using this option? i.e., How would you call this option in its default way, is it `--exclude-cols Metric Description` in full?

utils/split_config.py
```diff
@@ -0,0 +1,144 @@
+# NOTES
+#
+# Read utils/unified_config.yaml and split it into per gfx architecture per panel config files
```
xuchen-amd, line 3: Question on counter/metric definitions. Does this mean when updating/editing any counter metric definitions, we will need to edit the unified_config.yaml?

xuchen-amd (commented):
Leaving some questions. The split looks good!

src/rocprof_compute_soc/analysis_configs/gfx90a/1200_local_data_share_lds.yaml
feizheng10, line 1: 1200_local_data_share_lds.yaml. May I know the reason of file name change?
I'd like to keep the original one

utils/split_config.py
vedithal-amd: The only thing developers should touch is `unified_config.yaml` , no other yaml file should be touched!

After updating `unified_config.yaml` we should run `split_config.py` which will auto-generate all other yaml files.

This is what we decided previously :)

src/argparser.py
vedithal-amd: I think there is a misunderstanding

By default 'Metrics Description' column is excluded from the analysis output in cli mode

To prevent such an exclusion, we need to explictly say not to exclude anything by passing `--exclude-cols` option with no arguments.

I think the line `Defaults to 'Metric Description'` threw us off, will fix this

src/rocprof_compute_soc/analysis_configs/gfx90a/1200_local_data_share_lds.yaml
vedithal-amd: The title of panel id `1200` in the unified config says `Local Data Share (LDS)`, I have not changed this, this is how it was previously.

When we autogenerate per gfx per panel configs for analysis, we replace the parenthesis and spaces with underscore, hence the autogenerated file name is `1200_local_data_share_lds.yaml`

I believe this is a better name than the previous name `1200_lds.yaml`, why do you think its not a good file name?

If you absolutely dont like the new file name, then we will have to update the title of panel id `1200` in unified config to `LDS` instead of `Local Data Share (LDS)`

But I dont understand whats the big deal here, the new name is more clear to the customer

utils/split_config.py
vedithal-amd: Clarification, when i say any other yaml file, I meant analysis config yaml files and docs/data/metrics_description.yaml not the mi gpu spec yaml file

There is an explicit test case which ensures that the modification timestamp of unified_config.yaml should be less than any other yaml file, this way developers wont forget to run the split_config.py script

However, if a developer directly updates the auto generate yaml file and tries to commit that, we dont have any test case to prevent this. Perhaps a hash based solution would work, where we store the file hash of autogenerated config files and ensure their file hash dont change. I will add this test

vedithal-amd: Of course customers can update analysis config files by hand but they will have to disable autogen test case to ensure test pass.

But the developer should never touch analysis config files unless they are testing something

tests/test_gpu_specs.py
vedithal-amd: I had to make this change to ensure ctest passes

xuchen-amd: This change will restrict where ctest can be run by looking for `src/rocprof-compute`.

utils/split_config.py
xuchen-amd: imo we will need another solution (unless we already have something in mind) to allow easier modification of `unified_config.yaml` when needed (not part of this PR). Fishing through a file of 3k+ lines to modify something is not optimal.

tests/test_gpu_specs.py
vedithal-amd: I was able to run ctests from `build` directory and as far as I know ctest is usually run from inside the `build` directory.
Running pytest also worked from project root directory.

Could you provide a reproducer on how this change is breaking where ctest can be run

utils/split_config.py
vedithal-amd: I think this was considered to be an acceptable pain point in our discussions as the whole point of creating a unified config file was to avoid developers having to update multiple files

docs/sphinx/requirements.txt
```diff
@@ -226,6 +226,8 @@ sphinx-design==0.6.1
     # via rocm-docs-core
 sphinx-external-toc==1.0.1
     # via rocm-docs-core
+sphinx-jinja==2.0.2
+    # via rocm-docs-core
```
peterjunpark: Hi @vedithal-amd,
`requirements.txt` shouldn't be edited manually
```
# This file is autogenerated by pip-compile with Python 3.10
# by the following command:
#
#    pip-compile requirements.in
```
Could you add `sphinx-jinja==2.0.2` to [requirements.in](https://github.com/ROCm/rocprofiler-compute/blob/develop/docs/sphinx/requirements.in) then [pip-compile](https://pypi.org/project/pip-tools/) it?

vedithal-amd: Thanks for the heads up, this is done now

src/rocprof_compute_soc/analysis_configs/gfx90a/1200_local_data_share_lds.yaml
feizheng10: Sorry for the late response.
"Local Data Share (LDS)" is good for text in official documentation description.
IMO, "local_data_share_lds" is duplicate desc from programmer view. "shmem" at NV land and "lds" at AMD are conventionally accepted standard. Here, we should keep "lds" or  ""local_data_share" only.

feizheng10:
I like the part you populate to the rst! Still thinking about the yaml part.

feizheng10:
"--exclude-cols" seems not the best name for me. Not sure for others. Not a big deal:)

src/argparser.py
vedithal-amd: @feizheng10 , i will rename as --include-cols and update the documentation

feizheng10:
> "--exclude-cols" seems not the best name for me. Not sure for others. Not a big deal:)


--exclude-cols implicitly to say: reveal some or all hidden columns.  Is it exactly what we want say here? 
I can see the literal extensibility a bit from the table design side for the future through... 

I'd suggest to change it to something like:
--show-description, --toggle-description,
or the short version of those,
--show-desc, --toggle-desc.

And, we need to make sure "--show-desc" works only in table view.

src/rocprof_compute_soc/analysis_configs/gfx906/1500_address_processing_unit_and_data_return_path_ta_td.yaml
```diff
@@ -0,0 +1,299 @@
+# AUTOGENERATED FILE. Only edit for testing purposes, not for development. Generated from utils/unified_config.yaml. Generated by utils/split_config.py
```
feizheng10: "address_processing_unit_and_data_return_path_ta_td", this is not a good practice for naming conversion.
If we really want to unify those, let's keep the whole full name only, like address_processing_unit_and_data_return_path. 

And lower case L1 and L2 are bad idea that engineers complained in the past.

src/rocprof_compute_soc/analysis_configs/gfx940/1000_compute_units_instruction_mix.yaml
```diff
@@ -0,0 +1,381 @@
...
+      plain: The total number of fused multiply-add instructions operating on 64-bit
+        floating-point operands issued to the VALU per normalization unit.
+      unit: Instructions per normalization unit
+    F64-Trans:
+      plain: The total number of transcendental instructions (such as sqrt) operating
+        on 64-bit floating-point operands issued to the VALU per normalization unit.
+      unit: Instructions per normalization unit
+    Conversion:
+      plain: "The total number of type conversion instructions (such as converting\
+        \ data to or from F32\u2194F64) issued to the VALU per normalization unit."
```
feizheng10, line 325: "\u2194" seems to me not for plain text?

feizheng10 (approved):

PR #727, opened 2025-06-04
added progress printout in rocprof-compute profile

jamessiddeley-amd:
Resolves the lack of progress feedback in rocprof-compute profile. Introduces a printout (e.g., "[Run X/Y]") before each application execution. Includes time estimation past the first run. 
E.g: [Run 1/13][Approximate profiling time left: 1 minute 41 seconds]...]

Example output:
prog_profiler_output


src/rocprof_compute_profile/profiler_base.py
```diff
@@ -329,6 +329,22 @@ def run_profiling(self, version: str, prog: str):
         console_debug(
             "profiling", "performing profiling using %s profiler" % self.__profiler
         )
+        
+        # helper function to format time for readability
+        def format_time(seconds):
```
feizheng10: Maybe move to util?

feizheng10 (approved):

feizheng10 (approved):

src/rocprof_compute_profile/profiler_base.py
```diff
@@ -430,11 +445,15 @@ def run_profiling(self, version: str, prog: str):
                 )
             else:
                 console_error("Profiler not supported")
+            end_run_time = time.time()
+            run_duration = end_run_time - start_run_time
```
vedithal-amd: Why reinvent the wheel, look where we are calling run_prof, we are already calculating profiling time over there, it is better to use that time

src/rocprof_compute_profile/profiler_base.py
```diff
@@ -362,7 +363,21 @@ def run_profiling(self, version: str, prog: str):
         input_files = glob.glob(self.get_args().path + "/perfmon/*.txt")
         input_files.sort()
 
-        for fname in tqdm(input_files, disable=disable_tqdm):
+        total_runs = len(input_files)
+        total_time_so_far = 0
+        avg_run_time = 0
+
+        for i, fname in enumerate(tqdm(input_files, disable=disable_tqdm)):
```
vedithal-amd: I can see `disable_tqdm` is set to True, if we are not using tqdm lets remove the function call and import statement.
Check if we are using tqdm anywhere else

vedithal-amd (changes requested):
Can you copy paste an example output in the description of the PR

vedithal-amd (approved):
LGTM

PR #728, opened 2025-06-05
fixed long kernel names cut off in --kernel-names option

jamessiddeley-amd:
Fixed kernel names being cut off when dumped out into legend in PDF using --kernel-names option with generating standalone roofline plot. Changed PDF generation to wrap long kernel names as well as use a table with 2 columns (symbol number and kernel name) rather than the current unreliable scatter plot implementation.


feizheng10 (approved):

cfallows-amd:
Discussed offline- minor adjustment to symbol output in PDF, simple change and will be quickly retested.

cfallows-amd (approved):
With offline discussions and result images shared, looks good to me, after the conflict in roofline.py file is resolved- that line can be removed as it has been relocated in the file since. Just do a quick retest to make sure it runs.

PR #732, opened 2025-06-08
TUI improvement 

feizheng10:
* Fixed broken arg for TUI (came from merging conflict)
* Updated TUI version and copyright  in some source file
* Minor label changes in TUI
* Enable pc sampling in TUI

### DIFF
--- a/src/rocprof_compute_tui/config.py
+++ b/src/rocprof_compute_tui/config.py
@@ -1,3 +1,27 @@
+##############################################################################bl
+# MIT License
+#
+# Copyright (c) 2025 Advanced Micro Devices, Inc. All Rights Reserved.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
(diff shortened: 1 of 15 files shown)


src/rocprof_compute_tui/config.py
```diff
@@ -6,7 +30,7 @@
 
 # Application settings
 APP_TITLE = "ROCm Compute Profiler TUI"
-VERSION = "3.1.0"
+VERSION = "3.2.0"
```
vedithal-amd, line 33: Why have a separate version for TUI when we already have a version for while rocprofiler compute

src/utils/parser.py
```diff
@@ -1284,7 +1282,9 @@ def load_pc_sampling_data(workload, dir, file_prefix, sorting_type):
             pc_sampling_method = "host_trap"
 
     if pc_sampling_method == None:
-        console_error("PC sampling: can not find %s " % csv_file_path)
+        console_warning(
+            "PC sampling: can not dect pc sampling method without %s " % csv_file_path
```
vedithal-amd: there is a typo, it should be `detect`

vedithal-amd (approved):
Just need to fix a small typo, other than that LGTM

cfallows-amd (approved):
LGTM, on top of comments from Vignesh for review.

src/rocprof_compute_tui/config.py
feizheng10: No, we shouldn't. We should pop up the string from the root VERSION to here in the build system, otherwise just manually.

src/utils/parser.py
feizheng10: Thanks! Fixed with [d098824](https://github.com/ROCm/rocprofiler-compute/pull/732/commits/d098824752d872ccd8062688e86a62c17875f163)

PR #733, opened 2025-06-08
Bugfix for rocprofiler sdk interface not working in MI 200

vedithal-amd:


feizheng10 (approved):
Does it happen to MI200 only?

vedithal-amd:
> Does it happen to MI200 only?

It happens in MI 200 and below `"mi50", "mi60", "mi210", "mi250", "mi250x"`

PR #737, opened 2025-06-09
Add roofline cli_generate_plot method

cfallows-amd:
Add option to print out roofline plot in terminal using plotext

Takes in one datatype and prints the roofline for said datatype.
Cannot stack types into one plot due to colour/symbol limitations and overall plot line and legend messiness- becomes illegible.
Cannot do subplots for each datatype (if we were to pass in a list of types) and return one single plot due to known issue with log math errors in plotext src.

Tested with the lines below that runs through all datatypes from run parameters to generate multiple plots. Each call to method is for one datatype, and returns plt.build() of one single plot. Tested on MI300X using recent rocm staging build.
for dtype in self.__run_parameters['roofline_data_type']:
            plt._utility.write(self.cli_generate_plot(dtype))

Attached image shows example of debug prints for all parameters, plus full plot.
Screenshot 2025-06-09 133000

### DIFF
--- a/src/roofline.py
+++ b/src/roofline.py
@@ -591,7 +591,7 @@ class Roofline:
         else:
             console_warning("No MFMA measurement available for {}".format(dtype))
 
-        # # Plot Application AI,  todo: show kernel name
+        # Plot Application AI
         for cache_level in cache_hierarchy:
             key = "ai_" + cache_level.lower()
             if key in self.__ai_data:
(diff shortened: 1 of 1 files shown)


src/roofline.py
```diff
@@ -414,6 +421,214 @@ def generate_plot(self, dtype, fig=None) -> go.Figure():
 
         return fig
 
+    # ----------------------------------------
```
feizheng10: Could you update the func comments as I suggested before?

feizheng10 (approved):
LG. Just one minor suggestion.

PR #738, opened 2025-06-09
Revert default interface to rocprof v1

vedithal-amd:
* Add deprecation warning for rocprof v1 / v2 / v3 interfaces to use rocprofiler-sdk interface


feizheng10 (approved):

vedithal-amd:
The default was changed to rocprofv3 in https://github.com/ROCm/rocprofiler-compute/pull/748

PR #740, opened 2025-06-10
Fix load_kernel_top arg for GUI analyze mode

cfallows-amd:
--gui option for analyze mode failing due to missing arg in load_kernel_top call in pre_processing
Manually tested --gui option to confirm fix. This change does not affect regular analyze run, only gui mode option.


vedithal-amd (approved):

feizheng10 (approved):
Thanks!

PR #744, opened 2025-06-11
Remove rocscope related code and add deprecation warning for mongo db usecase

vedithal-amd:
* Remove rocscope related code
* Remove gfx perfmon build which is not being used
* Add deprecation warning for mongo db usecase


feizheng10:
rocscope related code can be removed, for sure.
For Grafana and Mongo database, some unused util or helper might be removed. But the whole solution can only be deprecated after TUI solution is good enough.

vedithal-amd:
> rocscope related code can be removed, for sure. For Grafana and Mongo database, some unused util or helper might be removed. But the whole solution can only be deprecated after TUI solution is good enough.

@feizheng10 , is there anything in this PR delta which you dont want me to remove?

feizheng10:
> > rocscope related code can be removed, for sure. For Grafana and Mongo database, some unused util or helper might be removed. But the whole solution can only be deprecated after TUI solution is good enough.
> 
> @feizheng10 , is there anything in this PR delta which you dont want me to remove?

I'm wondering src/utils/db_connector.py is stil in using when some one import data to Mongo DB?

vedithal-amd:
> > > rocscope related code can be removed, for sure. For Grafana and Mongo database, some unused util or helper might be removed. But the whole solution can only be deprecated after TUI solution is good enough.
> > 
> > 
> > @feizheng10 , is there anything in this PR delta which you dont want me to remove?
> 
> I'm wondering src/utils/db_connector.py is stil in using when some one import data to Mongo DB?

In this PR, I have removed the command line option for mongo db functionality such as db import.
No one should be importing mongo db data because we decided to deprecate it.
If we know anyone who is still using mongo db with rocprofiler compute we should tell them to move away from that

vedithal-amd:
Per my conversation with @feizheng10  we cannot remove the mongo db code until TUI is productized.
Instead I will add a deprecation warning for mongo db usage

vedithal-amd:
@feizheng10  and @peterjunpark could you please review this, thanks!

feizheng10 (approved):

PR #747, opened 2025-06-12
Fix: Add explicit includes for assert.h and thread in sample/vcopy.cpp

jamessiddeley-amd:
The HIP/CLR change e3cb5399c removed transitive inclusion   of standard headers like assert.h from hip_runtime.h.   This caused build failures in rocprof-compute samples.  

This commit explicitly includes <assert.h> and <thread>   in vcopy.cpp to resolve potential missing definitions.

### DIFF
--- a/sample/vcopy.cpp
+++ b/sample/vcopy.cpp
@@ -25,6 +25,8 @@
 */
 
 #include "hip/hip_runtime.h"
+#include <assert.h>
+#include <thread>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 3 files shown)


vedithal-amd (approved):
LGTM given cmake build is passing in https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=34068&view=logs&j=4356c74d-32e5-5e9d-6ea9-de34e14b3950&t=3881389b-a86e-54be-9826-43d19be781ba

PR #748, opened 2025-06-12
Change default rocprof to rocprofv3

vedithal-amd:
* Change default rocprof backend interface to rocprofv3

* Add MI 350 support in documentation

* Added known issue that MI 100 profiling will not work unless rocprofv1 is explicitly opted in

* Remove MI 50 soc gfx python class since MI 50 is not supported


vedithal-amd:
This is in continuation to https://github.com/ROCm/rocprofiler-compute/pull/738

feizheng10 (approved):

PR #750, opened 2025-06-12
fix broken test with --cols options

ywang103-amd:


cfallows-amd (approved):
Assuming ctest tests failing are now passing. LGTM in code.

feizheng10 (approved):

PR #751, opened 2025-06-12
Fix: Add explicit includes for assert.h and thread in sample files

jamessiddeley-amd:
https://github.com/ROCm/rocprofiler-compute/pull/747


vedithal-amd (approved):
Approved due to previous linked PR

PR #752, opened 2025-06-13
Fix test_gpu_specs.py

vedithal-amd:


cfallows-amd (approved):
Assuming ctest has been run locally to confirm fix. Src code change LGTM. Thanks

PR #754, opened 2025-06-13
Ignore certain workflows when changes are only for .md, docs, and docker

jbonnell-amd:

## Related Issue

- [ ] Bug Fix
- [ ] Cherry Pick
- [X] Continuous Integration
- [ ] Documentation Update
- [ ] Feature
- [ ] Optimization
- [ ] Refactor
- [ ] Other (please specify)

## Technical Details

When adding the pull request template, we noticed that the build pipelines were still running even for small modifications to markdown files. To avoid running these long workflows on minor changes, I have added `paths-ignore` parameters as shown below:

- *.md
  - Any .md file in the root
- .github/**/*.md
  - Any .md file within the .github directory
- doc/**
  - Any files within the doc directory
- docker/**
  - Any files within the docker directory

This means that if a push or PR happens and the only files changed are in these directories, it will ignore running the workflow. These parameters have been added to the following workflows:

- rhel-8.yml
  - On pr and push
- tarball.yml
  - On pr only
  - On push, it currently only triggers on `amd-mainline` and `release/**` branches so I thought it would make sense to keep this as-is instead of adding extra rules
- ubuntu-jammy.yml
  - On pr and push

Let me know if there are any concerns about these proposed changes, or if there are any additional exclusions I could add for these build workflows. 

- [ ] Yes

- [ ] Yes

- [ ] Yes


.github/workflows/formatting.yml
```diff
@@ -12,6 +12,11 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+
+  call-workflow-passing-data:
```
jbonnell-amd: This is calling the following workflow from ROCM/rocm-docs-core:

https://github.com/ROCm/rocm-docs-core/blob/develop/.github/workflows/linting.yml

It will perform a linting check on all markdown files, as well as perform a spellcheck. This is currently causing a lot of failures on both the markdown linting (should be easy to fix but would require a lot of changes) and spellcheck (we likely need to add/update wordlist.txt to include technical terms), so I will remove these changes for now. If we want to implement this, we can have a separate issue created/PR that would just add this along with all of the necessary fixes to make the pipelines pass.

.github/workflows/rhel-8.yml
```diff
@@ -6,8 +6,18 @@ name: RHEL 8
 on:
   push:
     branches: [ amd-mainline, amd-staging, release/**, develop ]
+    paths-ignore:
```
jbonnell-amd, line 9: I tried looking into ways to reduce having to copy and paste the same list of `paths-ignore` for every workflow, but it seems that sharing variables between YAML files is not something that is natively supported. I looked into some potential solutions:

1. Create an anchor
    - This is the YAML equivalent of variables, but can only be accessed within a single file and not across different ones
    - GitLab documentation seems to suggest sharing across files is possible
    - https://docs.gitlab.com/ci/yaml/yaml_optimization/#anchors
2. Use GitHub environment variables
    - Can be stored either within a single workflow or across a repository
    - Probably not a good idea for something that should be more visible, since storing it in a repository means it will be difficult to access
    - https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables

For now, I left it as-is but am curious if either of these solutions seem appealing to implement. Either way, something to consider for future usage.

feizheng10 (approved):

PR #755, opened 2025-06-15
Update roofline bins for rocm7 changes

cfallows-amd:

## Related Issue

- [ ] Bug Fix
- [ ] Cherry Pick
- [ ] Continuous Integration
- [ ] Documentation Update
- [ ] Feature
- [ ] Optimization
- [x] Refactor
- [x] Other (please specify)
Rebuild roofline binaries

## Technical Details

Hip changes causing roofline binaries built on 6.4 to fail. Needed rebuild for each distro on top of rocm7 hip changes. Roof bins once again labelled by rocm version, which is factored in again when looking for roofline bins during profiling. No change to roofline functionality.
Built using docker images of base OS distributions, with rocm7.0 staging builds. Roofline bins manually run in each docker image. Removed runpaths from bins and compared rocm6 to rocm7 bins for each distro that the glibc versions are the same.

- [ ] Yes

- [ ] Yes

- [ ] Yes


xuchen-amd (approved):

src/utils/utils.py
```diff
@@ -1230,7 +1233,7 @@ def detect_roofline(mspec):
     else:
         console_error("roofline", "Cannot find a valid binary for your operating system")
 
-    target_binary = {"distro": distro}
+    target_binary = {"rocm_ver": rocm_ver, "distro": distro, "path": None}
```
vedithal-amd: Why do we need `path`, I dont see it being used anywhere?

src/utils/utils.py
```diff
@@ -1201,6 +1203,7 @@ def detect_roofline(mspec):
         if path(rooflineBinary).exists():
             console_warning("roofline", "Detected user-supplied binary")
             return {
+                "rocm_ver": "override",
```
vedithal-amd: Why do we need `path`, I dont see it being used anywhere?

vedithal-amd (changes requested):
The issue still persists, check the Azure pipeline, `rocprofiler_compute_test_gfx942_staging` -> `View more details on Azure Pipeline` ->  `rocprofiler-compute Test` -> search for `roofline-ubuntu22_04-rocm7`

https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=34439&view=logs&j=723463d1-36b1-5dff-ad94-31990425871e&t=e1c5f8bc-5ed0-5dbc-826e-aaf61d056d45&l=1786

It is still trying to find rocm 6 version shared library and emits the following error even though the roofline binary says rocm 7

`roofline-ubuntu22_04-rocm7: error while loading shared libraries: libamdhip64.so.6: cannot open shared object file: No such file or directory`

cfallows-amd:
PR on pause while we sort out build/test issues, blocked by other rocm stack components.

src/utils/utils.py
cfallows-amd: The problem was the inconsistency of the return value- a target_binary of different size and keys depending on whether user supplied ROOFLINE_BIN or was using default name and location.  We should be returning a consistent dict so that it is clear for users and devs what the expected values are in both cases. Based on the code I don't expect to be caught in a bad state, but we should be consistent anyways for ctest coverage additions.

src/utils/utils.py
cfallows-amd: In mibench() method a call to detect_roofline() that has a user-supplied ROOFLINE_BIN gets propagated back to mibench through 'path' value. Otherwise mibench uses 'rocm_ver' and 'distro' values to create the default path.

cfallows-amd:
Used Build #16379 (Jul 7, 2025, 12:04:55 PM) to compile rocm-amdgpu-bench and create and test the roofline binaries on the respective base OS docker images. Complete manual testing and ctest runs performed on ubuntu22.04 using same #16379 build that contains hip version 7 changes. Confirmed on rhel8 image that roofline with rocm7 is no longer supported- rocm7 starts at rhel9 now.
Note- test_utils.py runs are currently failing multiple related and unrelated tests- these tests are brand new and need to be assessed for initial pass-ability. ip_blocks is failing due to SPI output, and misc failing due to functionality change for roof pdf outputs (tets is no longer valid)- adding changes and currently testing for these two failing tests, they are unrelated to this PR.

vedithal-amd:
I have tested that the rocm 7 binaries work with HIP 7!

vedithal-amd (approved):
LGTM

CHANGELOG.md
```diff
@@ -78,6 +78,7 @@ Full documentation for ROCm Compute Profiler is available at [https://rocm.docs.
   * Total FLOPs (consider fp6 and fp4 ops)
 * Update Dash to >=3.0.0 (for web UI)
 * Change when Roofline PDFs are generated- during general profiling and --roof-only profiling (skip only when --no-roof option is present)
+* Update Roofline binaries- OS distribution support minimum is now Ubuntu22.04, RHEL9, and SLES15SP6
```
xuchen-amd: white space between `binaries` and `-`?

cfallows-amd: minor rework after discussion with Xuan- add clarity that this is roofline minimums.

PR #764, opened 2025-06-20
Fix counter collection for MI 350

vedithal-amd:

## Related Issue

- [x] Bug Fix
- [ ] Cherry Pick
- [ ] Continuous Integration
- [ ] Documentation Update
- [ ] Feature
- [ ] Optimization
- [ ] Refactor
- [ ] Other (please specify)

## Technical Details

* Fix TCC counter collection bug
* Fix accumulate counter collection
* Fix formatting

- [ ] Yes

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/utils/utils.py
+++ b/src/utils/utils.py
@@ -34,12 +34,14 @@ import selectors
 import shutil
 import subprocess
 import sys
+import tempfile
 import time
 from collections import OrderedDict
 from itertools import product
 from pathlib import Path as path
 
 import pandas as pd
+import yaml
 
 import config
 from utils.logger import (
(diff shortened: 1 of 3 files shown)


src/utils/utils.py
```diff
@@ -853,6 +894,10 @@ def run_prof(
         )
     )
 
+    # Delete temporary files
```
feizheng10, line 897: We have to delete those explicitly ?

feizheng10 (approved):
LG. Would you add comments somewhere: the reason to do so in case sdk fixed in the future.

src/utils/utils.py
vedithal-amd: It's not good to leave temporary files/folders in user's system.
They get cleaned up automatically by system but if we clean them explicitly that's a better experience for the user.

vedithal-amd:
> LG. Would you add comments somewhere: the reason to do so in case sdk fixed in the future.

sdk team suggested to use this workaround and refused to add any fix to rocprofv3

PR #790, opened 2025-07-04
Enable rocpd output format with rocprofiler sdk

vedithal-amd:

## Related Issue

- [ ] Bug Fix
- [ ] Cherry Pick
- [ ] Continuous Integration
- [ ] Documentation Update
- [x] Feature
- [ ] Optimization
- [ ] Refactor
- [ ] Other (please specify)

## Technical Details
Enable rocpd output format with rocprofiler sdk
* Add `rocpd` choice for `--format-rocprof-output` option
* Add rocpd_data.py which defines SQL queries to extract data from rocpd database
* Use sqlite3 package to read the database
* Add `--retain-rocpd-output` option in profile mode to retain raw
  rocpd database
* Add deprecation warning to say `--format-rocprof-output rocpd` will be
  default in future release

For rocpd output:
* Use only `pmc_perf.csv` instead of reading individual coll_level results csv files
* Post process csv files using pandas in analysis mode instead of profile mode
* Use ACCUM counters instead of SQ_ACCUM_PREV_HIRES

Others:
* Add test cases for rocpd output format
* Fix code formatting issues
* Update CHANGELOG

- [x] Yes

- [ ] Yes

- [x] Yes


src/argparser.py
```diff
@@ -307,6 +307,13 @@ def validate_block(value):
         default="/opt/rocm/lib/librocprofiler-sdk.so",
         help="\t\t\tSet the path to rocprofiler SDK library.",
     )
+    profile_group.add_argument(
+        "--retain-rocpd-output",
```
feizheng10, line 284: Is this option reserved for future?

src/utils/parser.py
```diff
@@ -314,7 +314,14 @@ def build_eval_string(equation, coll_level):
     # use .get() to catch any potential KeyErrors
     s = re.sub(r"raw_pmc_df\['(.*?)']", r'raw_pmc_df.get("\1")', s)
     # apply coll_level
-    s = re.sub(r"raw_pmc_df", "raw_pmc_df.get('" + coll_level + "')", s)
+    if config.get("format_rocprof_output") == "rocpd":
+        # Replace SQ_ACCUM_PREV_HIRES with coll_level_ACCUM then ignore coll_level df
```
feizheng10, line 318: Could you add explanation in the comments why we need to do so? for code reading in the future

src/utils/utils.py
```diff
@@ -839,7 +845,29 @@ def run_prof(
 
     results_files = []
 
-    if rocprof_cmd.endswith("v2"):
+    if format_rocprof_output == "rocpd":
+        if rocprof_cmd == "rocprofiler-sdk" or rocprof_cmd.endswith("v3"):
+            # Write results_fbase.csv
+            rocpd_data.convert_db_to_csv(
```
feizheng10, line 843: Any specific reason to convert to csv? Why not read_sql_query directly?

src/utils/rocpd_data.py
```diff
@@ -0,0 +1,93 @@
...
+    Merge counters across unique dispatches from the input dataframe and return processed dataframe.
+    """
+    # Only import pandas if needed
+    import pandas as pd
+
+    data = list()
+    # Group by unique kernel and merge into a single row
+    for _, group_df in df.groupby(
+        [
+            "Dispatch_ID",
```
feizheng10, line 62: I'd suggest to put this kind of definition in a centrelized place because we are using those a few places.
In the past, rocprof change those without notification...It is even better to check the columns existing or not in advanced.

src/utils/utils.py
vedithal-amd: We only care about counters_collection view which is in KBs in size. The whole database is 50 MB in size and has multi dimensional counter data per dispatch, but we only care about aggregate counter data per dispatch.

There is an option to retain the raw large database file on the harddisk instead of deleting it, in case people want to do something with it.

Think of this as intermidate format absolutely private to rocprofiler-compute, post analysis will then read this csv for post analysis which contains the same data raw data as the database.

src/argparser.py
vedithal-amd: Yes this will be useful in future if we want some other multidimensional data from database

src/utils/parser.py
vedithal-amd: Sure, will add more comments.

To clarify, in rocpd there is no concept of `coll_level` and `SQ_ACCUM_PREV_HIRES`. We only have `<coll_level>_ACCUM` counters in rocpd which are collected using the special definition provided by us.

src/utils/rocpd_data.py
vedithal-amd: Can you suggest where we can store this?

src/utils/utils.py
feizheng10: We care about aggregate counter data with user selected range of dispatches or the default all, right?

src/utils/rocpd_data.py
feizheng10: In the past, we put common defs in schema.py. 
Or somewhere root for profiler_base/v3.

src/utils/utils.py
vedithal-amd: Thats right

src/utils/file_io.py
```diff
@@ -194,12 +194,11 @@ def create_df_kernel_top_stats(
 
 @demarcate
 def create_df_pmc(
-    raw_data_root_dir, nodes, spatial_multiplexing, kernel_verbose, verbose
+    raw_data_root_dir, nodes, spatial_multiplexing, kernel_verbose, verbose, config
```
feizheng10, line 196: It is debatable passing profile_config all the way down. Not a big deal for short term.

feizheng10 (approved):
As we discussed, it is ok as short-term solution.
Some minor comments. Approved. Thanks!

PR #795, opened 2025-07-08
Update MI100 Counters

darren-amd:

## Related Issue

- [ ] Bug Fix
- [ ] Cherry Pick
- [ ] Continuous Integration
- [ ] Documentation Update
- [x] Feature
- [ ] Optimization
- [ ] Refactor
- [ ] Other (please specify)

- [ ] Yes

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/rocprof_compute_soc/profile_configs/gfx908_counter_defs.yaml
+++ b/src/rocprof_compute_soc/profile_configs/gfx908_counter_defs.yaml
@@ -1439,7 +1439,7 @@ rocprofiler-sdk:
       - gfx908
       block: TCC
       event: 34
-  - name: TCC_EA_WRREQ_LEVEL
+  - name: TCC_EA0_WRREQ_LEVEL
     description: The sum of the number of EA write requests in flight. This is primarily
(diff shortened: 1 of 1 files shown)


src/rocprof_compute_soc/profile_configs/gfx908_counter_defs.yaml
```diff
@@ -1385,7 +1385,7 @@ rocprofiler-sdk:
       - gfx908
       block: TCC
       event: 27
-  - name: TCC_EA_WR_UNCACHED_32B
+  - name: TCC_EA0_WR_UNCACHED_32B
```
vedithal-amd, line 1388: Why are we changing counter names here?
If we change the name here we would also have to update the formula of derived metrics in analysis config yaml files.
We would only like to make any MI 100 related changes if something is broken, is that the case here?

vedithal-amd: If we change these counter names, we would also need to update the metric formulas for MI 100 in analysis config files, which seems unnecessary at this point.

vedithal-amd:
Its better to just add descriptions, which i have done in https://github.com/ROCm/rocprofiler-compute/pull/834.
Closing this PR.

PR #806, opened 2025-07-11
rocm-smi deprecation warning

xuchen-amd:

## Related Issue

- [ ] Bug Fix
- [ ] Cherry Pick
- [ ] Continuous Integration
- [ ] Documentation Update
- [ ] Feature
- [ ] Optimization
- [ ] Refactor
- [x] Other (Deprecation Warning)

## Technical Details
Add warning message for users still relying on rocm-smi.

- [ ] Yes

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/utils/specs.py
+++ b/src/utils/specs.py
@@ -164,15 +164,16 @@ def generate_machine_specs(args, sysinfo: dict = None):
     rocm_smi_compute_partition_pattern = r"Compute Partition:\s*(\S+)"
 
     vbios = search(vbios_pattern, amd_smi_output)
-    # 1. get compute partition from amd-smi
+
+    # Non-high-frequency System Patterns
+    # Option 1. Try amd-smi (default)
     compute_partition = search(compute_partition_pattern, amd_smi_output)
     console_debug(f"amd-smi compute partition: {compute_partition}")
-    # 2. get compute partition from rocm-smi
+    # Option 2. Try rocm-smi (DEPRECATION NOTE: rocm-smi removal in ROCm 7.1)
     if compute_partition is None:
         compute_partition = search(
             rocm_smi_compute_partition_pattern, rocm_smi_compute_partition_output
         )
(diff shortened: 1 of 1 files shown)


src/utils/specs.py
```diff
@@ -172,6 +172,14 @@ def generate_machine_specs(args, sysinfo: dict = None):
         compute_partition = search(
             rocm_smi_compute_partition_pattern, rocm_smi_compute_partition_output
         )
+        # DEPRECATION NOTE: rocm-smi removal in ROCm 7.1
```
vedithal-amd: You are still using amd-smi in the step after this right?
rocm-smi should be moved one step below such that all amd-smi related things are tried first.

vedithal-amd (changes requested):

src/utils/specs.py
xuchen-amd: The plan is to only add deprecation message for 7.0, and remove rocm-smi in 7.1, this is for 7.0 cherry pick.

xuchen-amd: as for the order, the amd-smi with `ACCELERATOR` tag is specific for system of high-frequency variants, which is its own search pattern.

vedithal-amd: Thanks for the context, so the users of high frequency gpus will still see this rocm-smi deprecation message even though amd-smi works for them perfectly. This will make them confused as to how can they migrate to amd-smi even though amd-smi works for them.

xuchen-amd: No, rocm-smi won't work on High Frequency systems.

rocm-smi is a back up for amd-smi on non-HF systems. And amd-smi on HF system is its on category.
Updating the comments section to make this clear.

vedithal-amd: I see now, as long as the users of HF systems dont see that deprecation message, its good to go.

vedithal-amd (approved):

feizheng10 (approved):

PR #811, opened 2025-07-15
Fix roofline block print during CLI output

cfallows-amd:

## Related Issue

- [X] Closes #SWDEV-542013

- [X] Bug Fix
- [ ] Cherry Pick
- [ ] Continuous Integration
- [ ] Documentation Update
- [ ] Feature
- [ ] Optimization
- [ ] Refactor
- [ ] Other (please specify)

## Technical Details

* Fix roofline panel in CLI analyze stage. Former output- roof chart prints every time even with block filer, current output- roof chart only prints when no filter, or if filter contains block 4.
* If roof was not profiled in profile stage, user gets following already-existing info message to let them know
"   INFO [roofline] /dockerx/src/rocprofiler-compute/install/bin/workloads/occupancy2/MI308X/roofline.csv does not exist"
but added another in the actual block output to explicitly let users know there is incomplete roofline data as a console error log that does not exit, but just prints and continues through other blocks as usual.
* Fixed check for baseline comparison- was originally skipping all hidden panels every time without checking if more than one path was provided, now skips only if user gives more than one workload path. Roofline panel is not printed when doing a baseline compare- roofline methods are not equipped at this time to handle more than one workload path. We skip roofline panel block.

Tested manually on ubuntu24.04 with Build #16387 and latest develop rocprof-compute branch + PR, using rocprofv3. Trialed various combinations of --block filter in analyze to ensure roof chart displays only when requested, and without a roofline.csv to confirm no exit errors occur when roof was never profiled prior.

- [ ] Yes

- [ ] Yes

- [ ] Yes


feizheng10 (approved):
LG

PR #817, opened 2025-07-17
Improve baseline comparison logic

vedithal-amd:

## Related Issue

- [x] Bug Fix
- [ ] Cherry Pick
- [ ] Continuous Integration
- [ ] Documentation Update
- [ ] Feature
- [ ] Optimization
- [ ] Refactor
- [ ] Other (please specify)

## Technical Details

* Do not force unsupported metrics to be specified in older gpu
  architectures as None

* Remove metrics which are explicitly set to None

* Update CHANGELOG

* Fix analysis configuration to fix baseline comparisons across all gpu
  architectures
    * Add missing 1812 section for gfx908
    * Add missing 1812 section for gfx90a

* Baseline comparison should only show common metrics

LDS Stats section for gfx950 (MI 350 series)

{EFED311E-FA90-45D4-8890-8BE26D951EE4}

LDS Stats section for gfx942 (MI 300 series)

{66929835-8BBE-47AD-AD8D-C6C708AEA0EE}

LDS Stats section in baseline comparison mode between gfx950 (first workload) and gfx942 (second workload).
Note that Metric IDs correspond to first workload.

{D076CBD4-2410-4072-8646-5689BBC272C1}

- [x] Yes
Validated baseline comparison working across all gpu architectures

- [ ] Yes

- [x] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,8 @@ Full documentation for ROCm Compute Profiler is available at [https://rocm.docs.
   * OS distribution support minimum for roofline feature is now Ubuntu22.04, RHEL9, and SLES15SP6
 * Improve analysis block based filtering to accept metric id level filtering
   * This can be used to collect individual metrics from various sections of analysis config
+* CLI analysis mode baseline comparison will now only compare common metrics across workloads and will not show Metric ID
+  * Remove metrics from analysis configuration files which are explicitly marked as empty or None
 
 ### Optimized
 
--- a/src/rocprof_compute_soc/analysis_configs/gfx908/0200_system-speed-of-light.yaml
+++ b/src/rocprof_compute_soc/analysis_configs/gfx908/0200_system-speed-of-light.yaml
@@ -32,12 +32,6 @@ Panel Config:
             peak: (((($max_sclk * $cu_per_gpu) * 64) * 2) / 1000)
             pop: None # No perf counter
             tips:
-          MFMA FLOPs (F8):
-            value: None # No HW module
-            unit: GFLOP/s
-            peak: None # No HW module
-            pop: None # No HW module
-            tips:
           MFMA FLOPs (BF16):
             value: None # No perf counter
             unit: GFLOP/s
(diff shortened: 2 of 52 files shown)


feizheng10 (changes requested):
1. Please separate gfx906 deprecation out of this pr.
2. Keep block IDs is important. When we are talking about "baseline comparison", the 1st run is used as baseline reference, which implies all the info from the 1st run are the baseline including block IDs.

feizheng10:
And just for internal maintain/design purpose, we  have meta info in the comments(temporally) for the reasons why we don't have the metrics available. It is better to provide solution when removing those.

vedithal-amd:
> 1st run is used as baseline reference, which implies all the info from the 1st run are the baseline including block IDs.

My concern is, does it make sense to compare a metric when its only available in one run?
Lets say 1st run has Metric1 = 10, Metric2 = 10 and 2nd run does not have Metric1 but Metric2 = 5.
In above case, does it make sense to only compare Metric2 which is common?
Why will the user want to see Metric1 in baseline comparison when it is not there in run 2? Why cant the user just do analysis of run 1 to see Metric1?

> Please separate gfx906 deprecation out of this pr.

OK

> And just for internal maintain/design purpose, we have meta info in the comments(temporally) for the reasons why we don't have the metrics available. It is better to provide solution when removing those.

We can backport the metrics from MI350 config to older GPUs when someone ask for it in a different PR.
There is no point in having None metrics when you can look at the metric information in newer MI350 GPU config.

vedithal-amd:
> Please separate gfx906 deprecation out of this pr.

This will be tracked in https://github.com/ROCm/rocprofiler-compute/pull/819

vedithal-amd:
> Keep block IDs is important. 

There is no such thing as block IDs, I think you are referring to Metric ID and i dont think that adds any value to the analysis report. Metric ID will still be shown in non baseline comparison mode. Only in baseline comparison mode its not able to show Metric ID due to metrics being different.

vedithal-amd:
As discussed on call, we can add Metric ID from first workload during baseline comparison.

feizheng10 (approved):
LGTM. Thanks!

PR #818, opened 2025-07-17
Live attach detach

ywang103-amd:


src/utils/utils.py
```diff
@@ -291,65 +292,79 @@ def capture_subprocess_output(
...
+
     while process.poll() is None:
         # Wait for events and handle them with their registered callbacks
         events = selector.select()
         for key, mask in events:
             callback = key.data
             callback(key.fileobj, mask)
 
-    # Get process return code
+    input_thread.join(timeout=1)
```
feizheng10, line 354: Any reason for timeout=1?

vedithal-amd (commented):
Please add documentation and CHANGELOG and add documentation team to this PR for review.

src/argparser.py
```diff
@@ -118,6 +118,24 @@ def omniarg_parser(
...
     )
     profile_group.add_argument("--target", type=str, default=None, help=argparse.SUPPRESS)
+    profile_group.add_argument(
+        "--pid",
+        type=str,
+        dest="pid",
+        metavar="",
+        default=None,
+        required=False,
+        help="\t\t\tenable attach/detach mode and set its process id.",
```
vedithal-amd, line 129: Better help message would be `Process id to be attached for profiling`

src/argparser.py
```diff
@@ -118,6 +118,24 @@ def omniarg_parser(
...
+    )
+    profile_group.add_argument(
+        "--attach-duration-msec",
+        type=str,
+        dest="attach_duration_msec",
+        metavar="",
+        default=None,
+        required=False,
+        help="\t\t\tWhen --pid is used, sets the amount of time in milliseconds the profiler will be attached before detaching. When unset, the profiler will wait until Enter is pressed to detach",
```
vedithal-amd, line 138: Does `seconds` make more sense here?

src/argparser.py
```diff
@@ -118,6 +118,24 @@ def omniarg_parser(
...
+        "--pid",
+        type=str,
+        dest="pid",
+        metavar="",
+        default=None,
+        required=False,
+        help="\t\t\tenable attach/detach mode and set its process id.",
+    )
+    profile_group.add_argument(
+        "--attach-duration-msec",
```
vedithal-amd, line 132: If using seconds, then rename to `--attach-duration` everywhere

src/rocprof_compute_profile/profiler_base.py
```diff
@@ -319,14 +319,15 @@ def pre_processing(self):
                 )
             self.__args.remaining = " ".join(self.__args.remaining)
         else:
-            console_error(
-                "Profiling command required. Pass application executable after -- at the end of options.\n\t\ti.e. rocprof-compute profile -n vcopy -- ./vcopy -n 1048576 -b 256"
-            )
+            if not self.__args.pid:
```
vedithal-amd, line 324: Use `elif` to avoid indentation

vedithal-amd: What if the user provides both ``--pid`` and ``-- <command>``?
We should add check for this and error out with appropriate error message saying that `command to be profiled should not be given when using the option --pid`

src/argparser.py
vedithal-amd: `args.remaining` should be set to empty string in argparser if ``--pid`` is given

src/rocprof_compute_profile/profiler_base.py
```diff
@@ -319,14 +319,15 @@ def pre_processing(self):
...
+                console_error(
+                    "Profiling command required. Pass application executable after -- at the end of options.\n\t\ti.e. rocprof-compute profile -n vcopy -- ./vcopy -n 1048576 -b 256"
+                )
 
         gen_sysinfo(
             workload_name=self.__args.name,
             workload_dir=self.get_args().path,
-            app_cmd=self.__args.remaining,
+            app_cmd=self.__args.remaining if self.__args.remaining else "",
```
vedithal-amd, line 332: As mentioned above, just set ``args.remaining`` to empty string in argparser when ``--pid`` is given.
This way you dont have to handle this everywhere.

src/rocprof_compute_profile/profiler_rocprof_v3.py
```diff
@@ -41,7 +41,10 @@ def __init__(self, profiling_args, profiler_mode, soc, supported_archs):
         )
 
     def get_profiler_options(self, fname, soc):
-        app_cmd = shlex.split(self.get_args().remaining)
+        app_cmd = (
+            shlex.split(self.get_args().remaining) if not self.get_args().pid else ""
```
vedithal-amd, line 47: Like mentioned above, handle it in one place.

src/rocprof_compute_profile/profiler_base.py
```diff
@@ -423,6 +424,15 @@ def run_profiling(self, version: str, prog: str):
             console_log("profiling", "Current input file: %s" % fname)
 
             options = self.get_profiler_options(fname, self._soc)
+
+            # Only 1-run case is permitted for attach/detach
+            if "--pid" in options and total_runs > 1:
+                console_error(
+                    "Attach/Detach can only work with 1 run of application. Current configuration needs {} times of replay of application".format(
```
vedithal-amd, line 437: Better help message would be...
Its better to use fstrings.
``f"Cannot attach process for profiling as number of required application replays {total_runs} is greater than 1"``

src/rocprof_compute_profile/profiler_base.py
```diff
@@ -451,6 +461,7 @@ def run_profiling(self, version: str, prog: str):
             else:
                 console_error("Profiler not supported")
             total_profiling_time_so_far += actual_profiling_duration
+
```
vedithal-amd, line 470: Remove whitespace

src/rocprof_compute_profile/profiler_rocprof_v3.py
```diff
@@ -60,6 +63,19 @@ def get_profiler_options(self, fname, soc):
             "--output-format",
             self.get_args().format_rocprof_output,
         ]
+
+        if self.get_args().pid:
+            args = args + [
```
vedithal-amd, line 70: use `args.append`

src/rocprof_compute_profile/profiler_rocprof_v3.py
```diff
@@ -60,6 +63,19 @@ def get_profiler_options(self, fname, soc):
...
+                "--pid",
+                self.get_args().pid,
+            ]
+
+            if self.get_args().attach_duration_msec:
+                args = args + [
+                    "--attach-duration-msec",
+                    self.get_args().attach_duration_msec,
```
vedithal-amd, line 78: If you use seconds, you can do `str(self.get_args().attach_duration * 1000`

src/argparser.py
```diff
@@ -118,6 +118,24 @@ def omniarg_parser(
...
+        type=str,
```
vedithal-amd, line 133: Time should be integer, right?

src/rocprof_compute_profile/profiler_rocprof_v3.py
```diff
@@ -77,8 +93,11 @@ def get_profiler_options(self, fname, soc):
                     dispatch.append(f"{int(dispatch_id) + 1}")
         if dispatch:
             args.extend(["--kernel-iteration-range", f"[{','.join(dispatch)}]"])
-        args.append("--")
-        args.extend(app_cmd)
+
+        if not self.get_args().pid:
```
vedithal-amd, line 99: You have only updated rocprofv3 interface which is going to be deprecated soon.
Please update rocprofiler-sdk interface as well.

src/utils/utils.py
```diff
@@ -299,50 +301,62 @@ def capture_subprocess_output(
...
             bufsize=1,
+            stdin=subprocess.PIPE,
             stdout=subprocess.PIPE,
             stderr=subprocess.STDOUT,
             universal_newlines=True,
             env=new_env,
         )
     )
 
-    # Create callback function for process output
```
vedithal-amd, line 308: Lets not remove comments which can improve readability of code

src/utils/utils.py
```diff
@@ -299,50 +301,62 @@ def capture_subprocess_output(
...
     buf = io.StringIO()
 
     def handle_output(stream, mask):
         try:
-            # Because the process' output is line buffered, there's only ever one
-            # line to read when this function is called
             line = stream.readline()
+            if not line:
```
vedithal-amd, line 317: We should skip `handle_output` in attach/dettach mode

src/utils/utils.py
```diff
@@ -299,50 +301,62 @@ def capture_subprocess_output(
...
+            except Exception:
+                pass
+
+    input_thread = threading.Thread(target=forward_input, daemon=True)
+    input_thread.start()
+
     while process.poll() is None:
-        # Wait for events and handle them with their registered callbacks
-        events = selector.select()
+        events = selector.select(timeout=1)
```
vedithal-amd, line 349: Why timeout=1 is required?

src/utils/utils.py
```diff
@@ -299,50 +301,62 @@ def capture_subprocess_output(
...
             pass
 
-    # Register callback for an "available for read" event from subprocess' stdout stream
     selector = selectors.DefaultSelector()
     selector.register(process.stdout, selectors.EVENT_READ, handle_output)
 
-    # Loop until subprocess is terminated
+    def forward_input():
+        try:
+            for line in sys.stdin:
```
vedithal-amd, line 332: First, check if sys.stdin has anything, this way you ddont have to do try except

src/utils/utils.py
```diff
@@ -299,50 +301,62 @@ def capture_subprocess_output(
...
+                if process.poll() is not None:
+                    break
+                process.stdin.write(line)
+                process.stdin.flush()
+        except Exception:
+            pass
+        finally:
+            try:
+                process.stdin.close()
```
vedithal-amd, line 341: Why would this fail? try catch seems not necessary

src/utils/utils.py
```diff
@@ -299,50 +301,62 @@ def capture_subprocess_output(
...
+            except Exception:
+                pass
+
+    input_thread = threading.Thread(target=forward_input, daemon=True)
```
vedithal-amd, line 345: Threading is fine, but if you can do it using current `selector` approach thats even better

src/utils/utils.py
```diff
@@ -731,9 +747,16 @@ def run_prof(
 
     if using_v3():
         if rocprof_cmd == "rocprofiler-sdk":
+            if is_mode_live_attach:
```
vedithal-amd, line 749: This is wrong, `live attach/detach` should absolutely support `rocprofiler-sdk` because `rocprofv3` interface will be removed soon.

vedithal-amd (changes requested):
Process attachment should work with rocprofiler-sdk, not only with rocprofv3 since rocprofv3 interface will be removed soon.

src/utils/utils.py
ywang103-amd: Mark has told me that current attach/detach only support rocprofv3 and sdk needs to wait for their further releases

src/utils/utils.py
ywang103-amd: I tried selector and it could not forward the input. It 's probably because "getline" has an internal buffer and only trigger when the buffer is full.

jayhawk-commits:
Imported to ROCm/rocm-systems

src/argparser.py
ywang103-amd: I will keep ms as it's more accurate for the setup

src/argparser.py
ywang103-amd: it needs to be passed to rocprofv3 as argument in command line, so here should be a string

src/argparser.py
ywang103-amd: millisecond gives user more accurate control on when to exit and after which kernel to exit

PR #841, opened 2025-07-30
[PC Sampling]: quick fix for how to invoke rocprof

feizheng10:

## Related Issue

- [x] Closes # SWDEV-541054

- [x] Bug Fix
- [ ] Cherry Pick
- [ ] Continuous Integration
- [ ] Documentation Update
- [ ] Feature
- [ ] Optimization
- [ ] Refactor
- [ ] Other (please specify)

## Technical Details

The technical evaluation is: the profiling target app cmd string(in args.remaining) should be converted to list before calling rocprof. We have some good practice for other places in the code, https://github.com/ROCm/rocprofiler-compute/blob/c7b5a0f43aba4ec72b8bebcdacb579d9a52cf1bc/src/rocprof_compute_profile/profiler_rocprofiler_sdk.py#L43. We apply similar solution to this case as a quick fix. 
Discussed with @vedithal-amd, we could have better unified solution: apply a transformation function in argparser once(Commented in the code). Because of the time limitation, will do it later after next week. 

- [ ] Yes

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/rocprof_compute_profile/profiler_base.py
+++ b/src/rocprof_compute_profile/profiler_base.py
@@ -27,6 +27,7 @@ import glob
 import logging
 import os
 import re
+import shlex
 import shutil
 import time
 from abc import ABC, abstractmethod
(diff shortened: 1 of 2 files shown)


vedithal-amd (approved):
LGTM

PR #858, opened 2025-08-06
SWDEV-521636 - Add dependend script path to system path in rocprof-compute

rocm-devops:
In wheel environment, rocprof-compute in bin folder is not a soft link. Adding the dependency scripts path to system path will help in executing rocprof-compute from bin folder as well


jayhawk-commits:

